PR #5, opened 2017-05-11
Escape strings in jupyter-config-data json

seibs:
On Windows, following the CONTRIBUTING.md instructions in jupyterlab fails when loading JupyterLab on a json parse exception.  The jupyter-config-data JSON includes Windows paths where slashes are not escaped.  Calling json.dumps on all page_config values fixes this.


blink1073:
Thanks!  Patch release coming.

blink1073:
Released as v0.2.5.

PR #10, opened 2017-05-16
Remove notebook dependency

blink1073:
This will allow `pip install -U jupyterlab_launcher` to work without transitive deps.


blink1073:
Whoops, forgot that the handlers used apis from the notebook as well.

PR #19, opened 2017-09-14
Update settings handler to parse complex URL paths.

afshin:
Supports making API requests to the settings handler in with the new ID schema from https://github.com/jupyterlab/jupyterlab/pull/2936

(This is a backward compatible API change.)

### DIFF
--- a/jupyterlab_launcher/settings_handler.py
+++ b/jupyterlab_launcher/settings_handler.py
@@ -74,23 +74,6 @@ class SettingsHandler(APIHandler):
         self.set_status(204)
 
 
-def _path(root_dir, section_name, make_dirs = False):
-    """Parse the URL section name and find the local file system path."""
-
-    # Attempt to parse path, e.g. @jupyterlab/apputils-extension:themes.
-    try:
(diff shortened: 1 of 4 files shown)


jupyterlab_launcher/settings_handler.py
```diff
@@ -66,30 +57,55 @@ def patch(self, section_name):
...
+    # Attempt to parse path, e.g. @jupyterlab/apputils-extension:themes.
+    try:
+        package_dir, plugin = section_name.split(":")
+        parent_dir = os.path.join(root_dir, package_dir)
+        path = os.path.join(parent_dir, plugin + ".json")
+    # This is deprecated and exists to support the older URL scheme.
+    except Exception as e:
+        path = os.path.join(root_dir, section_name + ".json")
+
+    if make_dirs and parent_dir and not os.path.exists(parent_dir):
```
blink1073: parent_dir would only be defined if there was a `:` in the section name (need to pre-define it outside of the try block).

blink1073:
Verified against https://github.com/jupyterlab/jupyterlab/pull/2936 current master, thanks!

PR #25, opened 2017-10-19
Update settings handler to support JSON with comments.

afshin:
**[work in progress]**

This change will require us to jump to version `0.6.0`

### DIFF
--- a/jupyterlab_launcher/settings_handler.py
+++ b/jupyterlab_launcher/settings_handler.py
@@ -32,14 +32,14 @@ class SettingsHandler(APIHandler):
 
         schema = _get_schema(self.schemas_dir, section_name)
         path = _path(self.settings_dir, section_name, _file_extension)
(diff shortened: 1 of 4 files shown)


jupyterlab_launcher/settings_handler.py
```diff
@@ -31,13 +32,16 @@ def get(self, section_name):
 
         schema = _get_schema(self.schemas_dir, section_name)
         path = _path(self.settings_dir, section_name, _file_extension)
+        user = ''
         settings = dict()
 
         if os.path.exists(path):
-            with open(path) as fid:
+            with open(path) as file:
```
blink1073: `file` is a built in function name

afshin: D'OH!

blink1073:
Thanks!

PR #29, opened 2017-12-06
Allow an admin to override extension schema defaults.

afshin:
This PR allows an administrator to create a file called `overrides.json` in the JupyterLab application settings directory (which can be found by running `jupyter lab paths`. Each key in that file should be the name of an extension (*e.g.* ` "@jupyterlab/apputils-extension:themes"`) and the value of that key should be an object with individual settings that the administrator wants to override. For example, an `overrides.json` file might look like this:

```json
{
  "@jupyterlab/apputils-extension:themes": {
    "theme": "JupyterLab Dark"
  }
}
```

Using this functionality, any default value set by an extension author can be overridden.


afshin:
Fixes https://github.com/jupyterlab/jupyterlab/issues/3179

blink1073:
Nice, thanks!

PR #37, opened 2018-01-12
[WIP] Allow urls to be overridden, and add base_url to all urls

blink1073:
This will require some changes to JupyterLab, especially the handing of page url, so deferring until after the beta1 release.  This normalizes the behavior of all urls, and allows a custom spin to use their own url paths, which can be visible to the user (e.g. `/mylab/tree/`).

### DIFF
--- a/jupyterlab_launcher/handlers.py
+++ b/jupyterlab_launcher/handlers.py
@@ -176,6 +176,8 @@ class NotFoundHandler(LabHandler):
 def add_handlers(web_app, config):
     """Add the appropriate handlers to the web app.
     """
+    base_url = web_app.settings['base_url']
+
     # Normalize directories.
     for name in config.trait_names():
         if not name.endswith('_dir'):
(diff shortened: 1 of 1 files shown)


jupyterlab_launcher/handlers.py
```diff
@@ -176,17 +176,27 @@ def render_template(self, name, **ns):
...
     for name in config.trait_names():
         if not name.endswith('_dir'):
             continue
         value = getattr(config, name)
         setattr(config, name, value.replace(os.sep, '/'))
 
+    # Normalize urls.
+    for name in config.trait_names():
+        if not name.endswith('_url'):
+            continue
```
vidartf: What if someone was to add a config option `npm_registry_url` at some point? Wouldn't this test be too general?

blink1073: Good point, we should only handle urls that start with `/`.

blink1073:
Replaced by #51

PR #38, opened 2018-01-21
Fix the URL patterns: ujoin() creates slashes.

afshin:
cc: @vidartf

### DIFF
--- a/jupyterlab_launcher/handlers.py
+++ b/jupyterlab_launcher/handlers.py
@@ -183,10 +183,10 @@ def add_handlers(web_app, config):
         value = getattr(config, name)
         setattr(config, name, value.replace(os.sep, '/'))
 
-    # Set up the main page handler.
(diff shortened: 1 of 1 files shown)


blink1073:
Thanks!

vidartf:
Sorry for being slow here, I guess the ping doesn't hit when it's in an edit? Doesn't this make `/labrador` impossible again? I'd love that as a unit test 🐕

afshin:
@vidartf Tornado respects the order that you added regular expressions to the launcher, so if you wanted `/labrador` you cann still just add that somewhere *before* the `NotFoundHandler`. Is that sufficient to the need you anticipate?

vidartf:
Is it possible to make a given server extension load before another?

afshin:
Let's talk about this tomorrow, I'm not sure what the answer is.

blink1073:
They are loaded in alphabetical order

PR #44, opened 2018-06-08
move process utilities from jupyterlab to jupyterlab_launcher

ivanov:


ivanov:
Though strictly speaking this PR only add functionality to the jupyterlab_launcher package, it's easier to think of it as a major change for the downstream dependency in jupyterlab,. /cc @jasongrout

jasongrout:
Since it is a bit of a pain to do major releases of jupyterlab_launcher, I think I'd rather still call this 0.10.6 (which it technically should be) and just make jupyterlab depend on something 0.10.6 or later.

PR #48, opened 2018-07-04
Update /lab/api/workspaces REST endpoints.

afshin:
- [x] Support workspace listing
- [x] Add workspace listing test
- [x] Support workspace deletion
- [x] Add workspace deletion test

These changes are backward-compatible, so we can publish them as version `0.11.1`

### DIFF
--- a/jupyterlab_launcher/app.py
+++ b/jupyterlab_launcher/app.py
@@ -6,15 +6,13 @@
 from notebook.notebookapp import NotebookApp
 from traitlets import Unicode
 
-from .handlers import add_handlers, LabConfig
+from .handlers import add_handlers
 
 
 class LabLauncherApp(NotebookApp):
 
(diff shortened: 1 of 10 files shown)


jupyterlab_launcher/app.py
```diff
@@ -6,16 +6,14 @@
...
+from .handlers import add_handlers
 
 
 class LabLauncherApp(NotebookApp):
 
     default_url = Unicode('/lab',
                           help='The default URL to redirect to from `/`')
 
-    lab_config = LabConfig()
-
```
ian-r-rose: Why remove this? I don't understand where the `LabConfig` is coming from in this case.

afshin: You're right. I removed it testing stuff and forgot it should go back. Cheers!

PR #54, opened 2018-08-10
Fix PY2 handing of potential non-unicode paths being slugified.

afshin:
cf. this broken services test build output: https://travis-ci.org/jupyterlab/jupyterlab/jobs/414480761

services test build output


blink1073:
Thanks!

PR #55, opened 2018-08-11
Switch to Python 3.5+

blink1073:
cf https://github.com/jupyterlab/jupyterlab/issues/2096


SylvainCorlay:
Do we want to make `jupyter_server` the actual dependency?

blink1073:
Once it is released :wink:

PR #57, opened 2018-08-23
Change jupyterlab_launcher to jupyterlab_server, update workspaces and settings endpoints.

afshin:
* Changes the package name from `jupyterlab_launcher` to `jupyterlab_server` to be more in line with @SylvainCorlay's work on creating the `jupyter_server` package.
* Cleans up adding of handlers.
* Exports `slugify` and `WORKSPACE_EXTENSION` to allow creation of a workspaces CLI in JupyterLab (https://github.com/jupyterlab/jupyterlab/pull/5166)
* Adds a new endpoint to `GET` a listing of all settings in one request
* Settings REST responses now include `id` and `version`. The version information comes from the `package.json.orig` file that is copied into the schemas directory when a plugin is first installed.
* Exports `SETTINGS_EXTENSION` file extension.


blink1073:
Here we go...

PR #58, opened 2018-09-14
Don't search for missing assets

jasongrout:
One correction: we aren't guaranteed index.html is in the static directory, since it is a *template*, not a static asset built. Since we don't really know what is going to be in the static directory, we just remove the check altogether. An app that wants a check can implement it in a subclass when it knows what to look for.

Two simplifications: 1. use python 3 super(), and rewrite the js function using modern apis.

~~I still want to test this before merging....~~ (ready for review)


jasongrout:
Ready for review now...

PR #59, opened 2018-09-14
Make url defaults start with the page_url. 

jasongrout:
This makes it so that all urls are by default namespaced by the page_url.


blink1073:
Thanks!

blink1073:
Released as 0.2.

jasongrout:
Thanks!

PR #61, opened 2019-02-11
Update workspaces listing to include values.

afshin:
Changes the workspaces listing to include full workspaces to conform to the `IDataConnector` front-end interface.

This change is backward-incompatible and requires updating to version `0.3.0`.

### DIFF
--- a/jupyterlab_server/tests/test_workspaces_api.py
+++ b/jupyterlab_server/tests/test_workspaces_api.py
@@ -53,8 +53,9 @@ class WorkspacesAPITest(LabTestBase):
     def test_listing(self):
         # ID fields are from workspaces/*.jupyterlab-workspace
         listing = set(['foo', 'f/o/o/'])
+        output = set(self.workspaces_api.get().json()['workspaces']['ids'])
 
-        assert set(self.workspaces_api.get().json()['workspaces']) == listing
+        assert output == listing
 
     def test_put(self):
         id = 'foo'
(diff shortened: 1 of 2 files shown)


ian-r-rose:
Looks reasonable to me. I find the construction of `{ids: string[], values: IWorkspace[]}` to be a bit odd. Why is it this way instead of `{ [id: string]: IWorkspace }`?

afshin:
That's the `IDataConnector`'s `list()` method signature:
```typescript
list(query?: any): Promise<{ ids: V[]; values: T[] }>;
```
https://github.com/jupyterlab/jupyterlab/blob/master/packages/coreutils/src/interfaces.ts#L67

The reasoning was that while it might be trivial in JS/TS to use `Object.keys(...)`, it might be more accessible to simply offer the list of `ids` separately from the values. Do you think we should re-visit that interface?

afshin:
RE: `IDataConnector`, imagine a situation where your identifier is complex: `{ "uri": "foo://bar/baz", "token": "75213d68ac8198a79a3dd391908e4778" }`. Because we have the generic type `V` defining what identifiers look like, we have the flexibility of allowing data connectors that use composite IDs, but we cannot make JSON objects whose keys are of that composite type `V`.

Of course, that does not apply here and we *can* make this REST endpoint return a simple map that we then process on the client-side to conform to `IDataConnector#list`'s output. But that's the rationale.

ian-r-rose:
Ah, right. Thanks for the reminder.

PR #64, opened 2019-05-30
Friendly workspaces

afshin:
Return a workspaces for every well-formed workspace API request URL.

**Note**
I don't understand where the version commit comes from. I may have made a mistake here.


blink1073:
Ha!  0.3 is already released, so let's keep it.

PR #69, opened 2019-06-07
A 403.html file. Should close jupyterlab issue#6065

zerline:
I'm contributing this 403.html, out of trust for the specifications seen on the https://github.com/jupyterlab/jupyterlab/issues/6065 conversation. Yet I've not tested that, for it would be a little too long. Hope it's useful, though.


blink1073:
Thanks again!

PR #71, opened 2019-06-24
Use json5 to load settings files.

ian-r-rose:
This json5 implementation seems like is works okay -- it is written by someone on the Chromium team. It has all sorts of blinking disclaimers that it is extremely slow, but I don't think we need a particularly high-performance algorithm for this use-case.


blink1073:
Json5 isn't on conda-forge yet...

ian-r-rose:
Hrm, I was hoping to avoid re-implementing...

ian-r-rose:
This came up last week, where the server-side and client-side validators for the settings files gave different answers (particularly around ECMA identifiers-as-keys).

blink1073:
It is fairly straightforward to get something on conda-forge.  It doesn't have any dependencies.

blink1073:
I'm sick today, but here are the instructions: https://conda-forge.org/#add_recipe

ian-r-rose:
`pyjson5` is now on conda-forge, and using this branch prevents 500 errors when the user tries to save valid `JSON5`-formatted settings files.

jasongrout:
:tada:

PR #72, opened 2019-07-18
ensure the 'WHICH' command returns absolute path instead of relative path

tgrout:
WHICH command returns relative path when running in directory where the binary is and this causes errors in labextension install because we try to execute this relative path in a temporary directory.  Correction is making sure absolute path is returned (consistent with documentation).

### DIFF
--- a/jupyterlab_server/process.py
+++ b/jupyterlab_server/process.py
@@ -59,7 +59,7 @@ def which(command, env=None):
             raise ValueError(msg)
         raise ValueError('The command was not found or was not ' +
                 'executable: %s.' % command)
-    return command_with_path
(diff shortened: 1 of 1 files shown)


jasongrout:
Indeed, I don't see this on mac (shutil.which seems to give the absolute path), but I do see it on Windows:

```
>python
Python 3.7.4 (default, Aug  9 2019, 18:34:13) [MSC v.1915 64 bit (AMD64)] :: Anaconda, Inc. on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from shutil import which
>>> which('python')
'.\\python.EXE'
```

jasongrout:
Thanks!

PR #73, opened 2019-08-14
fix page config escaping

nicorikken:
Properly escape strings for JSON data, to resolve issue https://github.com/jupyterlab/jupyterlab/issues/7008 This PR takes the approach of explicitly converting all values to String to prevent issues with non-string in JSON output like boolean, and then printing them as JSON string. The outer quotes are stripped, as they are added in the index.html templating later on.

### DIFF
--- a/jupyterlab_server/handlers.py
+++ b/jupyterlab_server/handlers.py
@@ -92,7 +92,7 @@ class LabHandler(JupyterHandler):
 
         # JSON escape Python values in page_config
         for key, value in page_config.items():
-            page_config[key] = json.dumps(value)[1:-1]
+            page_config[key] = json.dumps(str(value))[1:-1]
 
         # Load the current page config file if available.
         page_config_file = os.path.join(settings_dir, 'page_config.json')
--- a/jupyterlab_server/handlers.py
+++ b/jupyterlab_server/handlers.py
@@ -90,6 +90,10 @@ class LabHandler(JupyterHandler):
             full_url = ujoin(base_url, getattr(config, name))
             page_config[full_name] = full_url
 
+        # JSON escape Python values in page_config
+        for key, value in page_config.items():
+            page_config[key] = json.dumps(value)[1:-1]
+
         # Load the current page config file if available.
         page_config_file = os.path.join(settings_dir, 'page_config.json')
         if os.path.exists(page_config_file):


nicorikken:
This PR still needs a version bump. I wasn't sure if I should add one in this PR.

nicorikken:
I've tested this PR on our zero-to-jupyterhub-k8s deployment in which the original issue occurred.

blink1073:
Nice!  I wonder if this approach allows us to pass objects as values. We currently have a hack in the client to allow us to pass lists.

blink1073:
cf https://github.com/jupyterlab/jupyterlab/blob/6625c6493fce1ad420c9e0850a20028f9f62a017/packages/coreutils/src/pageconfig.ts#L83

blink1073:
I tested this locally and it doesn't interfere with the standard behavior.  So if it fixes your issue, I say let's merge and release a patch of the server, which is a lot easier than releasing JupyterLab  :smile:

nicorikken:
@blink1073 regarding passing lists to the client: that might work. This PR handles every value as a String, as it was communicated to the frontend as Strings already. So if you send a Python List, it becomes a JSON escaped version of the python string conversion (`str(mylist)` or `mylist.__str__()`) Don't expect Javascript to properly parse this, as it should do the exact inverse the `str()` function call.

I think the better way forward is to convert the Python objects to JSON objects. That includes dictionaries, lists, integers and booleans. Now the jupyterlab_server renders them as strings (look at the index.html page source as posted in the linked issue). So instead of constructing the JSON object in the template (combining the page_config and some URLs) build the full `jupyter-config-data` object as a Python dict and render it out to JSON using something like [jsons.dumps()](https://jsons.readthedocs.io/en/latest/api.html#id3)

blink1073:
@jasongrout is working on a variant that uses jinja to do the escaping for us.

aksh-gupta:
Did we bump the version? I am on the latest version `2.23.0` and I still face this same issue.

fcollonval:
@aksh-gupta would you mind opening a new issue that feature was changed in https://github.com/jupyterlab/jupyterlab_server/pull/75? If you could also provide a reproducible test case it would be great so we can add a test on this.

PR #74, opened 2019-08-14
Add store ID to page config

saulshanabrook:


blink1073:
@vidartf, any objections?

vidartf:
The concept is good, its probably (?) ok to put it on the application object.

saulshanabrook:
> Its probably (?) ok to put it on the application object.

I only used it because of [a stack overflow response](https://stackoverflow.com/a/12243098):

> Application is the right object to store (semi-)persistent data. However, as suggested on other anwser, you should consider using some kind of database to store this data. 


I couldn't find any Tornado docs that mention this use case. Maybe @Zsailer would know?

Zsailer:
Like @vidartf, I *think* this is fine. I was trying to find a scenario like this in JupyterHub—thinking there might be some semi-persistent data case there—but was unsuccessful. 

Typically, all persistent data storage mechanisms are listed in "setting" of the tornado application and stored somewhere on the server, but if you don't need this data to persist after the server is stopped, this should work. The data *should* persist with each request...

blink1073:
Thanks @vidartf and @Zsailer, let's give this a shot!

blink1073:
Released as 1.0.2.

saulshanabrook:
@blink1073 Thank you!

PR #75, opened 2019-08-15
Escape template values in Jinja

jasongrout:
This escapes the values that go into the default template using the Jinja escaping functions.

It reverts #73 to prevent double-escaping.


blink1073:
Thanks!

PR #78, opened 2019-08-20
Add .json.orig files to sdists

toddrme2178:
There is currently only one, `package.json.orig`, but it is needed for the tests to pass.


blink1073:
Thanks!

blink1073:
I released v1.0.6 which has this file.

PR #79, opened 2019-10-26
JupyterLab Server as Server Extension

echarles:
This PR migrate the existing ServerApp and Handler as [Jupyter Server Extensions](https://jupyter-server.readthedocs.io/en/latest/frontends.html). https://github.com/jupyterlab/jupyterlab/pull/7416 must also be considered.

- [x] Migrate ServerApp and Handlers to Extensions
- [x] Migrate tests to use PyTest plugin `pytest_jupyter_server`
- [x] Create PyTest `pytest_jupyterlab_server` to use by other libraries
- [x] Pin to a released jupyter_server


jupyterlab_server/tests/schemas/@jupyterlab/shortcuts-extension/plugin.json
```diff
@@ -54,7 +54,7 @@
       "properties": {
         "command": { "default": "completer:invoke-notebook" },
         "keys": { "default": ["Tab"] },
-        "selector": { "default": ".jp-Notebook.jp-mod-editMode .jp-mod-completer-enabled" }
+        "selector": { "default": ".jp-jupyter_server.jp-mod-editMode .jp-mod-completer-enabled" }
```
vidartf: I'm pretty sure these should not change...

echarles:
Removed WIP from title. This PR is ready for review.

setup.py
```diff
@@ -42,7 +42,6 @@
     setup_args['install_requires'] = [
         'json5',
         'jsonschema>=3.0.1',
-        'notebook>=4.2.0',
```
vidartf: Requires jupyter_server instead?

vidartf:
Just a comment from a quick read through.

setup.py
echarles: Done. Thx Vidar.

echarles:
@Zsailer Great. This is back Green. Thx again for your help.

jupyterlab_server/app.py
```diff
@@ -4,18 +4,36 @@
...
+
+    @property
+    def app_namespace(self):
+        return self.extension_name
 
     default_url = Unicode('/lab',
                           help='The default URL to redirect to from `/`')
 
-    lab_config = LabConfig()
+    extension_url = '/lab'
```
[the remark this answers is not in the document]
echarles: Removed. Thx.

jupyterlab_server/pytest_plugin.py
```diff
@@ -0,0 +1,119 @@
...
+from jupyterlab_server.tests.utils import here
+from jupyterlab_server.app import LabServerApp
+
+def mkdir(tmp_path, *parts):
+    path = tmp_path.joinpath(*parts)
+    if not path.exists():
+        path.mkdir(parents=True)
+    return path
+
+app_settings_dir = pytest.fixture(lambda tmp_path: mkdir(tmp_path, 'app_settings'))
```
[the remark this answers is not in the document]
echarles: Yeah, I love them also.

bollwyvl:
I put that coverage threshold stuff in there: we're not in a great place, coverage-wise, but seems like a lot to give up. While we do have the report from that external vendor, now, having it as a real test shows we care.

Also re: installing jupyter_server: can this be done as a separate set of test excursions, where notebook is not installed? Tox can union coverage across multiple runs, I think, so i think it might be a good choice to support, while still keeping ci clean. 

More broadly, this project will probably need to accept people will continue to be running `notebook` for a while, if they have invested in custom tooling, but will still want the newer features.

Dropping compatibility will need to be clearly signaled with warnings, potentially all the way up to the UI, well before it is officially dropped (as part of a major version bump).

Totally astronaut architecting here: it may be worth considering, in a future effort, the treating of the dependency resolutions as entire separate _packages_, e.g. `jupyterlab_server_classic` and `jupyterlab_server_jserver`... and adopting the manager pattern, where the logic moved into a couple Managers, while the handlers and extensions lived in the appropriate packages. This would demonstrate it's _possible_ to run lab on top of different (async) servers, which has been expressed as being important to some downstreams that want to bring their own X for different parts of the stack. I could see this being clutch for themes and translations, where a large deployment would want to centralize these.

echarles:
> I put that coverage threshold stuff in there: we're not in a great place, coverage-wise, but seems like a lot to give up. While we do have the report from that external vendor, now, having it as a real test shows we care.

Sure, I removed that while trying to get CI green (it was not picking the def in setup.py as it was before) - Will put that back in place as soon as we have more CI stablity.

> Also re: installing jupyter_server: can this be done as a separate set of test excursions, where notebook is not installed? Tox can union coverage across multiple runs, I think, so i think it might be a good choice to support, while still keeping ci clean.

I was able to install jupyter_server from github by defining it in setup.py - Now itt does not work anymore... Sorry for not mentioning that (did that the jupyter_server gitter channel) - The plan is to depend soon on a jupyter_server release that @Zsailer will do.

> More broadly, this project will probably need to accept people will continue to be running notebook for a while, if they have invested in custom tooling, but will still want the newer features.

JupyterLab users will continue to be able to use notebook via the nbclassic package, see https://github.com/jupyterlab/jupyterlab/pull/7416

> Dropping compatibility will need to be clearly signaled with warnings, potentially all the way up to the UI, well before it is officially dropped (as part of a major version bump).

We have ensured full backward compatilbity via nbclassic (https://github.com/Zsailer/nbclassic) with log warning for traits to be moved to ServerApp and for existing server extensions. 

> Totally astronaut architecting here: it may be worth considering, in a future effort, the treating of the dependency resolutions as entire separate packages, e.g. jupyterlab_server_classic and jupyterlab_server_jserver... and adopting the manager pattern, where the logic moved into a couple Managers, while the handlers and extensions lived in the appropriate packages. This would demonstrate it's possible to run lab on top of different (async) servers, which has been expressed as being important to some downstreams that want to bring their own X for different parts of the stack. I could see this being clutch for themes and translations, where a large deployment would want to centralize these.

There is obviously room for evolution, (re)packaging and modularity once we can jlab as jserver extension in (which is the goal of this PR).

echarles:
I will also have a look at https://github.com/jupyterlab/jupyterlab_server/pull/79/files#diff-f63568f03d848bfe4c05811be269b915R90 (not sure why the json response comes back with a different layout)

jupyterlab_server/__init__.py
```diff
@@ -16,3 +16,11 @@
     'SETTINGS_EXTENSION',
     'WORKSPACE_EXTENSION'
 ]
+
+def _jupyter_server_extension_paths():
```
Zsailer, line 20: This function has been changed to `_jupyter_server_extension_points` in jupyter_server.

It will still work with this old name, but it would be good to update here.

echarles: Thx. Done for jupyterlab_server. I will roll that change in jupyterlab also

PR #82, opened 2020-03-07
Black and White Listings Handler

echarles:
This is a PR to implement Black and White listings for the Extension Manager as discussed in #7933 and #7967.

It adds a ListingsHandler that delivers the URI for the lists.

It must be seen in conjunction with https://github.com/jupyterlab/jupyterlab/pull/7989

### DIFF
--- a/jupyterlab_server/app.py
+++ b/jupyterlab_server/app.py
@@ -15,16 +15,16 @@ class LabServerApp(ServerApp):
     default_url = Unicode('/lab',
                           help='The default URL to redirect to from `/`')
 
-    blacklist_uri = Unicode('', config=True,
+    lab_config = LabConfig()
+
+    blacklist_uris = Unicode('', config=True,
         help="A list of comma-separated URIs to get the blacklist")
 
-    whitelist_uri = Unicode('', config=True,
+    whitelist_uris = Unicode('', config=True,
         help="A list of comma-separated URIs to get the whitelist")
 
     listings_refresh_ms = Integer(1000 * 60, config=True,
         help="The interval delay in milliseconds to refresh the lists")
-
-    lab_config = LabConfig()
 
     def start(self):
         add_handlers(self.web_app, self.lab_config)
--- a/jupyterlab_server/app.py
+++ b/jupyterlab_server/app.py
@@ -23,8 +23,8 @@ class LabServerApp(ServerApp):
     whitelist_uris = Unicode('', config=True,
         help="A list of comma-separated URIs to get the whitelist")
 
-    listings_refresh_ms = Integer(1000 * 60 * 5, config=True,
-        help="The interval delay in milliseconds to refresh the lists")
+    listings_refresh_seconds = Integer(60 * 60, config=True,
+        help="The interval delay in seconds to refresh the lists")
 
     listings_request_options = Dict({}, config=True,
         help="The optional kwargs to use for the listings HTTP requests \
--- a/jupyterlab_server/app.py
+++ b/jupyterlab_server/app.py
@@ -4,7 +4,7 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-from traitlets import Unicode, Integer
+from traitlets import Unicode, Integer, Dict
 
 from .server import ServerApp
 from .handlers import add_handlers, LabConfig
@@ -25,6 +25,10 @@ class LabServerApp(ServerApp):
 
     listings_refresh_ms = Integer(1000 * 60, config=True,
         help="The interval delay in milliseconds to refresh the lists")
+
+    listings_request_options = Dict({}, config=True,
+        help="The optional kwargs to use for the listings HTTP requests \
+            as described on https://2.python-requests.org/en/v2.7.0/api/#requests.request")
 
     def start(self):
         add_handlers(self.web_app, self.lab_config)
--- a/jupyterlab_server/app.py
+++ b/jupyterlab_server/app.py
@@ -16,13 +16,13 @@ class LabServerApp(ServerApp):
                           help='The default URL to redirect to from `/`')
 
     blacklist_uri = Unicode('', config=True,
-        help="The default URI to get the black list")
+        help="A list of comma-separated URIs to get the blacklist")
 
     whitelist_uri = Unicode('', config=True,
-        help="The default URI to get the white list")
+        help="A list of comma-separated URIs to get the whitelist")
 
     listings_refresh_ms = Integer(1000 * 60, config=True,
-        help="The interval time in milliseconds to refresh the lisings")
+        help="The interval delay in milliseconds to refresh the lists")
 
     lab_config = LabConfig()
 
--- a/jupyterlab_server/app.py
+++ b/jupyterlab_server/app.py
@@ -15,6 +15,12 @@ class LabServerApp(ServerApp):
     default_url = Unicode('/lab',
                           help='The default URL to redirect to from `/`')
 
+    blacklist_uri = Unicode('', config=True,
+        help="The default URI to get the black list")
+
+    whitelist_uri = Unicode('', config=True,
+        help="The default URI to get the white list")
+
     lab_config = LabConfig()
 
     def start(self):
--- a/jupyterlab_server/app.py
+++ b/jupyterlab_server/app.py
@@ -4,7 +4,7 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-from traitlets import Unicode
+from traitlets import Unicode, Integer
 
 from .server import ServerApp
 from .handlers import add_handlers, LabConfig
@@ -20,6 +20,9 @@ class LabServerApp(ServerApp):
 
     whitelist_uri = Unicode('', config=True,
         help="The default URI to get the white list")
+
+    listings_refresh_ms = Integer(1000 * 60, config=True,
+        help="The interval time in milliseconds to refresh the lisings")
 
     lab_config = LabConfig()
 
--- a/jupyterlab_server/app.py
+++ b/jupyterlab_server/app.py
@@ -23,7 +23,7 @@ class LabServerApp(ServerApp):
     whitelist_uris = Unicode('', config=True,
         help="A list of comma-separated URIs to get the whitelist")
 
-    listings_refresh_ms = Integer(1000 * 60, config=True,
+    listings_refresh_ms = Integer(1000 * 60 * 5, config=True,
         help="The interval delay in milliseconds to refresh the lists")
 
     listings_request_options = Dict({}, config=True,
--- a/jupyterlab_server/handlers.py
+++ b/jupyterlab_server/handlers.py
@@ -306,8 +306,7 @@ def add_handlers(web_app, config):
     whitelist_uris = settings_config.get('whitelist_uris', '')
 
     if (blacklist_uris) and (whitelist_uris):
-        raise Exception('Simultaneous blacklist_uris and whitelist_uris is not supported. Please define only one of those.')
-        # TODO(@echarles) Force exit here.
+        print('Simultaneous blacklist_uris and whitelist_uris is not supported. Please define only one of those.')
         import sys
         sys.exit(-1)
 
--- a/jupyterlab_server/handlers.py
+++ b/jupyterlab_server/handlers.py
@@ -322,7 +322,6 @@ def add_handlers(web_app, config):
     if whitelist_uris:
         ListingsHandler.whitelist_uris = set(whitelist_uris.split(','))
     
-    # TODO(@echarles) Get the log function and pass it to fetch_listings
     fetch_listings(None)
 
     handlers.append((listings_path, ListingsHandler, {}))
--- a/jupyterlab_server/handlers.py
+++ b/jupyterlab_server/handlers.py
@@ -15,7 +15,7 @@ from traitlets import HasTraits, Bool, Unicode, default
 from .server import JupyterHandler, FileFindHandler, url_path_join as ujoin
 from .workspaces_handler import WorkspacesHandler
 from .settings_handler import SettingsHandler
-from .listings_handler import ListingsHandler
+from .listings_handler import ListingsHandler, fetch_listings
 from .themes_handler import ThemesHandler
 
 # -----------------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


echarles:
@jasongrout @blink1073 @vidartf @saulshanabrook Wanted to have your inputs on the following for the listings URIs.

Rather than providing these via the settings, I provide them via an additional `ListingsHandler`. This brings more security as the user can not change them (my understanding is that there is no way to have read-only settings).

The `ListingsHandler` is [here](https://github.com/jupyterlab/jupyterlab_server/pull/82/commits/78b627c02d009bd6e75321697213945c63a0341c#diff-39080f855774d3c6602514c700244e1fR1-R16) and returns that [json file](https://github.com/jupyterlab/jupyterlab/pull/7989/commits/c7bfec98315ed8c721b99d396c5a749aab2dc715#diff-2eeb45d490094a9989b835a1e4518789R1-R8). That file will contain the default listings URI (a github repo) which can be overriden via Server traits, e.g. ` --LabServerApp.blacklist_uri=...`.

I have introduced in the build system a new `list` attribute as the json must be pulled in the built system in written in the staging (or dev folder). This is done [here](https://github.com/jupyterlab/jupyterlab/pull/7989/commits/7497d0f6687906f74fdb81ee9f2204d612fec860#diff-3a0ad6926f6378b6d925965f752f356aR284-R299).

Looking forward to your feedback on this. Does it sound reasonable to introduce these?

echarles:
@jasongrout @blink1073 @vidartf @saulshanabrook

Upon the previous comment, I also come with an additional one about the component (Browser or Server) that should fetch the listings.

For now, I have gone with option 1 as the NPM search is achieved by the Browser, so it sounded more aligned to also run the listings fetch by the Browser.

**Option 1 - Browser fetches the listings**

```
Browser            Server         Listings
-------            ------         --------
   |  get uris       |               |
   |----------------]|               |
   |                 |               |
   | return uris     |               |
   |[--------------- |               |
   |                 |               |
   |  get listings   |               |
   |--------------------------------]|
   |                 |               |
   | return Listings |               |
   |[------------------------------- |
```

**Option 2 - Server fetches the listings**

```
Browser            Server         Listings
-------            ------         --------
   |  get listings   |               |
   |----------------]| get listings  |
   |                 |--------------]|
   | return listings |               |
   |[--------------- |               |
```

Sub-Options:

2.1. Contact Listings on every Browser request
2.2. Contact Listings on Server startup + cache the listings
2.3. Option 1.2 + Schedule a refresh of the cache based on a parameter (e.g. every 5 minutes)

echarles:
Moving from `Option 1 - Browser fetches the listings` to `Option 2.2 - Server fetches the listings, contact Listings on Server startup + cache the listings`, see this [ListingsHandler](https://github.com/datalayer-contrib/jupyterlab-server/blob/ea7d12fb88ed025a8b3a327e4e690a9f1c76c3a5/jupyterlab_server/listings_handler.py).

blink1073:
I like the approach of using handlers.  However, we need a way to refresh the listing for the case of long-running servers that can't easily be restarted.

echarles:
Thx @blink1073 for your feedback.

I also think we need to refresh the listings, so it would be option `2.3. Option 2.2 + Schedule a refresh of the cache based on a parameter (e.g. every 5 minutes)`.

I will go this way and add a trait to make the refresh interval configurable.

jupyterlab_server/listings_handler.py
```diff
@@ -0,0 +1,119 @@
...
+        if len(ListingsHandler.whitelist_uris) == 0:
+            for w in listings['listings']['whitelist_uris']:
+                ListingsHandler.whitelist_uris.add(w)
+
+
+    def fetch_listings_uris(self):
+        self.log.info('Fetching Blacklist from {}'.format(ListingsHandler.blacklist_uris))
+        blacklist = []
+        for ListingsHandler.blacklist_uri in ListingsHandler.blacklist_uris:
+            r = requests.get(ListingsHandler.blacklist_uri)
```
timkpaine: discussed in call:
- be careful with blocking
- in enterprise external requests might be blocked without proxies (e.g. allow for arbitrary kwargs)

echarles: Thx Tim, good call. From https://requests.readthedocs.io/en/latest/user/advanced/#proxies a single additional trait, e.g. LabServerApp.http_proxy=http://user:pass@10.10.1.10:3128 would be enough. Any reason to foresee arbitrary kwargs?

timkpaine: what if my resource is protected by cookie-based auth? I might need to pass in cookies to the request.

echarles: So we can define a trait of type dictionary and use ` requests.request(method, url, **kwargs)`. https://2.python-requests.org/en/v2.7.0/api/#requests.request

I think it better maps what you were thinking to?

echarles: Added here https://github.com/jupyterlab/jupyterlab_server/pull/82/commits/0318022bfbda5dc00d2d56df21494f0494cf5d2e#diff-39080f855774d3c6602514c700244e1fR43

Resolving converstation

jupyterlab_server/app.py
```diff
@@ -17,6 +17,19 @@ class LabServerApp(ServerApp):
 
     lab_config = LabConfig()
 
+    blacklist_uris = Unicode('', config=True,
```
saulshanabrook: I see that the behavior of adding multiple URLs is basically a concat of all those lists. However, I wonder about just supporting one URL for now, to keep things as simple as possible. What was the thinking around supporting multiple?

jupyterlab_server/listings_handler.py
```diff
@@ -0,0 +1,73 @@
...
+                jitter=0.1
+                )
+            ListingsHandler.pc.start()
+
+    def get(self, path):
+        self.set_header('Content-Type', 'application/json')
+        if path == LISTINGS_URL_SUFFIX:
+            self.write(ListingsHandler.listings)
+        else:
+            self.write({})
```
saulshanabrook: What about just raising a 404 for other paths?

jupyterlab_server/handlers.py
```diff
@@ -292,6 +304,34 @@ def add_handlers(web_app, config):
         handlers.append((
             workspace_api_path, WorkspacesHandler, workspaces_config))
 
+    # Handle local listings.
+    if config.listings_dir:
+
+        settings_config = web_app.settings.get('config', {}).get('LabServerApp', {})
+        blacklist_uris = settings_config.get('blacklist_uris', '')
```
saulshanabrook: Since you specify defaults for these settings values above, can you just use the getitem and assume they will exist? Seems unnecessary to put defaults in two places.

jupyterlab_server/handlers.py
```diff
@@ -292,6 +304,34 @@ def add_handlers(web_app, config):
...
+        whitelist_uris = settings_config.get('whitelist_uris', '')
+
+        if (blacklist_uris != '') and (whitelist_uris != ''):
```
saulshanabrook:
```suggestion
        if blacklist_uris and whitelist_uris:
```

jupyterlab_server/handlers.py
```diff
@@ -292,6 +304,34 @@ def add_handlers(web_app, config):
...
+            raise Exception('Simultaneous blacklist_uris and whitelist_uris is not supported. Please define only one of those.')
+            # TODO(@echarles) Force exit here.
+            import sys
+            sys.exit(-1)
+
+        ListingsHandler.listings_refresh_ms = settings_config.get('listings_refresh_ms', 1000 * 60 * 5)
```
saulshanabrook: See here, same thing about defaults. Above the default seems to be set to 1 minute and here you set to default of 5 minutes (which I believe will never be used because above default is set?)

jupyterlab_server/listings_handler.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+LISTINGS_URL_SUFFIX='@jupyterlab/extensionmanager-extension/listings.json'
+
+
+def fetch_listings():
+    if len(ListingsHandler.blacklist_uris) > 0:
+        blacklist = []
+        for blacklist_uri in ListingsHandler.blacklist_uris:
+            print('Fetching blacklist from {}'.format(ListingsHandler.blacklist_uris))
```
saulshanabrook: I guess switch this to use logging?

jupyterlab_server/listings_handler.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class ListingsHandler(APIHandler):
+    """An handler that returns the listings specs."""
+
+    blacklist_uris = set()
+    whitelist_uris = set()
+    blacklist = []
+    whitelist = []
+    listings = ''
```
saulshanabrook: Seems like this is a duplicate of the data in the attributes above? Maybe just remove it and dynamically create that dict on each get? It makes it a bit harder to follow what's being set.

jupyterlab_server/handlers.py
```diff
@@ -154,6 +155,11 @@ class LabConfig(HasTraits):
                                 'schemas directory. If given, a handler will '
                                 'be added for settings.'))
 
+    listings_dir = Unicode('',
```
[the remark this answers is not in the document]
echarles: That is a rest of my initial implementation explained in https://github.com/jupyterlab/jupyterlab_server/pull/82#issuecomment-596064918

My initial idea was to provide the uris via a json file shipped in the typescript packages (a bit like the schemas). I have removed all that yesterday as it was giving too many options and did not bring value as we were now using the server to configure the behavior.

I am removing this

jupyterlab_server/handlers.py
```diff
@@ -163,6 +169,8 @@ class LabConfig(HasTraits):
 
     workspaces_url = Unicode(help='The url path of the workspaces handler.')
 
+    listings_url = Unicode(help='The theme url.')
```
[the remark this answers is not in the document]
echarles: Removing this, see next comment

jupyterlab_server/app.py
echarles: The goal is for an org to override in a way the default blacklist. At the beginning, we said that jupyterlab would publish a blacklist (that idea may have changed now). An org would then want to urgently blacklist or strip the user, still relying on the community maintained blacklist.

So with multiple uris you can rely on work done by others. You could also in a org have multiple departments that the user belongs to and those departements maintain their own blacklist/whitelist.

jupyterlab_server/listings_handler.py
echarles: Better. Done.

jupyterlab_server/handlers.py
echarles: jupyter traits, and specifically how they are implemented in jupyterlab / jupyterlab-server are  strange. The defaults are stated for the developer, but they do not work as expected in my understanding. I have spent a lot of time to debug and making sure the pieces were correctly setup, also discussed with Zach. The new server extensions will make it more clear, but for now, I propose to keep as it is implemented.

jupyterlab_server/handlers.py
echarles: Done. Thx.

jupyterlab_server/handlers.py
echarles: Updated the default. Same comment as above, the defaults are not processed as they could be.

jupyterlab_server/listings_handler.py
echarles: This method is accessed from 2 places. I have passed the logger from one place, but have issue to pass it from the other place (in `handlers.py`)

I have added a TODO and will double check for this. At worth, it will use print the first time it fetches the listings, and logger the next times.

jupyterlab_server/listings_handler.py
echarles: These are class level attributes. I can remove them but would need to write some guards in the various methods (init, fetch...) that access them. IMHO this would make the code less readable than having these fields present. As workaround I have added some comments for each of these fields. Would it work for you?

echarles:
Thx @saulshanabrook for your reviews and comments. I have commented on each of them and pushed the changes. I have a minor TODO (pass the log from the handlers.py).

jupyterlab_server/listings_handler.py
saulshanabrook: Ah sorry just mean remove the `listing` attribute not all the class attributes.

jupyterlab_server/app.py
saulshanabrook: Sounds good.

jupyterlab_server/listings_handler.py
saulshanabrook: Currently this logging stands out a bit since it uses a print statement:

Screen Shot 2020-03-18 at 7 48 31 AM

Hm... Looking at it briefly, I believe you could make this an instance method of the `ListingsHandler`, so that you can use `self.log` in the body of this function.

Then in the `ioloop.PeriodicCallback`, just pass in `self.fetch_listings` instead. Also maybe you could just call `self.fetch_listings()` in `initialize` instead of in `handlers.py`?

echarles: The latest push will only use print on start, then the periodic pull will use the log.

For now I am initializing via a class method to ensure the list are loaded on server start. At this point, the self.log is not available. After thee PeriodicCallback passes the self.log and the console will look good.

One option would be to not call the fetch_listing on server start, but this would penalize the first user as this first user would need for the HTTP request of the server to the listings servers.

I guess I could copy the code sections that create the logger when self.log is not available. Let me try that.

echarles: All good. Now always using logger.

Screenshot 2020-03-18 at 13 45 29

jupyterlab_server/listings_handler.py
echarles: `listing` attribute has been removed. thx.

jupyterlab_server/listings_handler.py
saulshanabrook: Great!

echarles:
I am trying to force exit the server launch but the exception does not do the job. Any tips on this? Thx.

@blink1073 @vidartf 

https://github.com/datalayer-contrib/jupyterlab-server/blob/2013e37cd0faaf900117df2d67f7c873db54be1d/jupyterlab_server/handlers.py#L308-L312

echarles:
I don't raise exception anymore but simply `sys.exit(-1)`

echarles:
@saulshanabrook CI is green. LGTM to merge.

echarles:
@saulshanabrook Thx for the merge. https://github.com/jupyterlab/jupyterlab_server/pull/82 should also be merged, a new `jupyterlab_server` be released, and then a PR on jupyterlab to bump jupyterlab_server dep (not defined to 'jupyterlab_server@ git+https://github.com/datalayer-contrib/jupyterlab-server@bw-list`

cc/ @blink1073

saulshanabrook:
Just merged that server release, but don't have permission to release on PyPi.

PR #85, opened 2020-04-11
Do not try to close the watch process file handle

blink1073:
This was causing the application to lock during shutdown when in watch mode.  Defer the file handle management to the OS during shutdown.


blink1073:
@meeseeksdev please backport to 1.0.x

PR #87, opened 2020-04-16
Change payload parsing to accept JSON5 as raw string in JSON payload

OneRaynyDay:
Related to PR https://github.com/jupyterlab/jupyterlab/pull/8225

### DIFF
--- a/jupyterlab_server/settings_handler.py
+++ b/jupyterlab_server/settings_handler.py
@@ -16,6 +16,9 @@ from .server import APIHandler, json_errors
 
 # The JupyterLab settings file extension.
 SETTINGS_EXTENSION = '.jupyterlab-settings'
+# Used for when JSON5 payload is wrapped in a JSON payload in the form:
+# {'raw': '...'}, where '...' is the JSON5 string.
+RAW_KEY = 'raw'
 
 
 def _get_schema(schemas_dir, schema_name, overrides):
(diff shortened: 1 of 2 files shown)


jupyterlab_server/settings_handler.py
```diff
@@ -16,6 +16,9 @@
 
 # The JupyterLab settings file extension.
 SETTINGS_EXTENSION = '.jupyterlab-settings'
+# Used for when JSON5 payload is wrapped in a JSON payload in the form:
```
jasongrout, line 19: Since this is only used one place, I think we should just hardcode it there. Otherwise we have to cross-reference this one variable definition to understand the code below.

OneRaynyDay: Sure, that's understandable. I'll hardcode it.

jupyterlab_server/settings_handler.py
```diff
@@ -242,19 +242,21 @@ def put(self, schema_name):
         if not settings_dir:
             raise web.HTTPError(500, settings_error)
 
-        raw = self.request.body.strip().decode(u'utf-8')
+        raw_payload = self.request.body.strip().decode(u'utf-8')
+        raw_settings = json.loads(raw_payload)['raw']
+        payload = json5.loads(raw_settings)
```
jasongrout, line 247: Since we are parsing user input - do you think we should wrap this in a try/except that raises a http 400 error like below? That would give user feedback, rather than just erroring out here with a server error.

OneRaynyDay: Sounds good, I can do that :)

OneRaynyDay: Not only did I change it to have exception handling, I also added tests for them :)

jasongrout:
Note that this change is backwards incompatible and must be coordinated with jupyterlab/jupyterlab#8225

blink1073:
Thanks again @OneRaynyDay!  I had to do a rebase in #114 to finish out this PR to get it in for the beta cycle of 2.0

PR #88, opened 2020-05-08
Kill the subprocess if it does not stop

blink1073:
Our `WatchHelper` subprocesses were not responding to `SIGTERM`

### DIFF
--- a/jupyterlab_server/process.py
+++ b/jupyterlab_server/process.py
@@ -119,7 +119,12 @@ class Process(object):
 
         # Wait for the process to close.
         try:
-            proc.wait()
+            proc.wait(timeout=1.)
+        except TimeoutError:
+            if os.name == 'nt':
+                os.kill(proc.pid, signal.SIGBREAK)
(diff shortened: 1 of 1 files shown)


jasongrout:
Thank you very much for figuring this out and fixing it!

blink1073:
@meeseeksdev please backport to 1.0.x

telamonian:
Do we know exactly why `WatchHelper` suddenly stopped exiting cleanly? All of a sudden I had to start chasing down and `kill <pid>`-ing a bunch of node processes whenever I launched `jupyter lab --watch`. My original sense was this was due to some change in a point release of webpack.

It's not essential, but it would be good to know the root cause in case eg we're missing some cleanup step somewhere for webpack.

blink1073:
I'm doing some exploration right now, see: #91 and https://github.com/jupyterlab/jupyterlab/pull/8433.  This issue is very hard to debug, despite years of experience dealing with subprocesses...

PR #91, opened 2020-05-17
Clean up terminate logic

blink1073:
Follow up to #88.  Our usage tests are timing out in JupyterLab.

In https://github.com/jupyterlab/jupyterlab/pull/8433 we set the kernel [`shutdown_wait_time`](https://github.com/jupyter/jupyter_client/blob/ea823fd84b4f5db0a050457b9b2cd7619e9a2d53/jupyter_client/manager.py#L66) to 1 sec, so we set the wait time here to 2 sec to give the kernels a chance to shut down.

### DIFF
--- a/jupyterlab_server/process.py
+++ b/jupyterlab_server/process.py
@@ -119,7 +119,7 @@ class Process(object):
 
         # Wait for the process to close.
         try:
-            proc.wait(timeout=1.)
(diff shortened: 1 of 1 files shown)


blink1073:
@meeseeksdev backport to 1.0.x

PR #94, opened 2020-07-03
Expose settings API to other handlers.

goanpeca:
Hi @afshin, would something like this be ok so other handlers can access the settings in a friendlier fashion?

On the localization work, I use the current locale set by the user and based on that I provide the available language in original and localized form (for the current local)

Thoughts?

### DIFF
--- a/jupyterlab_server/settings_handler.py
+++ b/jupyterlab_server/settings_handler.py
@@ -54,10 +54,9 @@ def _get_settings(settings_dir, schema_name, schema):
     Returns a tuple containing the raw user settings, the parsed user
     settings, and a validation warning for a schema.
     """
-
     path = _path(settings_dir, schema_name, False, SETTINGS_EXTENSION)
     raw = '{}'
-    settings = dict()
+    settings = {}
     warning = ''
     validation_warning = 'Failed validating settings (%s): %s'
     parse_error = 'Failed loading settings (%s): %s'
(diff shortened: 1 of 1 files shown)


goanpeca:
Ok, let me add some tests then :-p

afshin:
Is it possible to combine `_get_settings` and `get_settings` so that the handler uses the same exact logic? It might prevent an out of sync bug in the future.

goanpeca:
>Is it possible to combine _get_settings and get_settings so that the handler uses the same exact logic? It might prevent an out of sync bug in the future.

Yeah I think so. Will update the code.

goanpeca:
>Is it possible to combine _get_settings and get_settings so that the handler uses the same exact logic? It might prevent an out of sync bug in the future.

I did something else, get_settings is now used inside the handler and also as the API exposed to other handlers so now the handler and the API use the same `get_settings` (public) method. The private `_get_settings`, well is private and has a different signature so no risk getting out of sync now.

Still need to add some tests.

jupyterlab_server/settings_handler.py
```diff
@@ -54,10 +54,9 @@ def _get_settings(settings_dir, schema_name, schema):
     Returns a tuple containing the raw user settings, the parsed user
     settings, and a validation warning for a schema.
     """
-
     path = _path(settings_dir, schema_name, False, SETTINGS_EXTENSION)
     raw = '{}'
-    settings = dict()
+    settings = {}
```
goanpeca, line 59: This is faster to use than `dict`, so .. why not use it 🙃

blink1073:
Published as 1.2.0, JupyterLab will need an update: https://github.com/jupyterlab/jupyterlab/blob/master/setup.py#L154

PR #99, opened 2020-07-21
Add last_modified and created to settings and workspace API items

### LINKED ISSUE #98: Expose more file metadata about disk-backed content
As some new UI work for [workspaces](https://github.com/jupyterlab/jupyterlab/issues/6944) and [settings](https://github.com/jupyterlab/jupyterlab/issues/8148) is in-flight, it would be nice to offer some additional information about stuff stored on disk. The only thing that really jumps out at me (without adding a bunch of stuff like history/rollback), is the file-system level information, namely modified and created time. This would enable the UI to answer use cases like "find the settings i most recently changed " or "find the workspace I was working on last Tuesday"

It looks like a `workspace` already has a metadata field (`id` lives there), so adding a `last_modified` and `created` would be reasonable, presumably as ISO 8061 dates. 

```yaml
{"values: [{
  "data": {...},
  "metadata": {
    "id": "/lab",
    "created": "2020-07-21T00:19:20+0000",
    "last_modified": "2020-07-21T00:20:20+0000"
  }   
}]}
```

I guess these would be overwritten on import/export, with the on-disk time always winning.

The settings API response looks a little more complicated, and perhaps these could just be added to the top level of each member of the `settings` array.

```yaml
{
  "settings": [{
    "id": "@foo/bar",
    "raw": "...",
    "settings": {...},
    "schema": {...},
    "version": "0.1.0",
    "created": "2020-07-21T00:19:20+0000",
    "last_modified": "2020-07-21T00:20:20+0000"
  }]
}
```

... or `metadata` could be added, with these in it, and eventually extended... though the buck has to stop somewhere, and I don't imagine much but sadness if settings carried anything not validated by the schema, or how it would get populated... HTTP headers?

Of course, open to any other suggestions of what might be hidden in there that would help folks navigate this stuff!

bollwyvl:
- fixes #98

Here's a minimal proposal for adding the fields mentioned in #98 to the various API routes for workspaces and settings.

- A workspace may or may not have `metadata.created` or `.last_modified`, but if it appears, should always be an UTC ISO 8601 datetime.
  - since it gets `PUT` back, the user _could_ modify the value, and it would be written to disk, but then overwritten when read back out. seems a little odd, but still okay
- A settings item will _always_ have both `created` and `last_modified`, and can be a UTC ISO 8601 datetime, or `null` in the event that the user hasn't created a settings file (otherwise, there was actually no way to know, as it always returns at least `{}` in `raw`).
  - this value is never written to disk

These both use the newly shimmed `tz` in `server`, but as this doesn't provide a way to _parse_ UTC ISO strings, the tests now also use what `jsonschema` _would_ use, `strict-rfc3339`... and indeed, it, and a few other packages, as runtime dependencies might provide better conformance to the JSON Schema spec. 

Longer term... these handlers would probably benefit from some schema, which would go part of the way towards https://github.com/jupyterlab/jupyterlab/issues/6013.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: python
 python:
+    - 3.9-dev
+    - 3.8
     - 3.6
     - 3.5
 sudo: false
--- a/jupyterlab_server/server.py
+++ b/jupyterlab_server/server.py
@@ -22,6 +22,7 @@ try:
         GREEN_ENABLED, GREEN_OK, RED_DISABLED, RED_X
     )
     from notebook.utils import url_escape, url_path_join
+    from notebook import _tz as tz
 except ImportError:
     from jupyter_server.base.handlers import (                          # noqa
         APIHandler, FileFindHandler, json_errors, JupyterHandler
@@ -31,3 +32,4 @@ except ImportError:
     )
     from jupyter_server.serverapp import ServerApp, aliases, flags      # noqa
     from jupyter_server.utils import url_escape, url_path_join          # noqa
+    from jupyter_server import _tz as tz
--- a/jupyterlab_server/settings_handler.py
+++ b/jupyterlab_server/settings_handler.py
@@ -11,7 +11,7 @@ from jsonschema import ValidationError
 from jsonschema import Draft4Validator as Validator
 from tornado import web
 
-from .server import APIHandler, json_errors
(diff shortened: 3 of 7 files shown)


bollwyvl:
haha, no `fromisoformat` for me, py3.5/3.6!

PR #100, opened 2020-07-21
PR: Add CI with github

goanpeca:


.github/workflows/tests.yml
```diff
@@ -0,0 +1,45 @@
...
+  linux:
+    name: Linux py${{ matrix.PYTHON_VERSION }} tests
+    runs-on: ubuntu-latest
+    env:
+      CI: True
+      PYTHON_VERSION: ${{ matrix.PYTHON_VERSION }}
+    strategy:
+      fail-fast: false
+      matrix:
+        PYTHON_VERSION: ['3.5', '3.6', '3.8']
```
bollwyvl, line 22: Seems like they've already got 3.9: probably better to know sooner! Heck I'd throw at least an pypy in there, too!

.github/workflows/tests.yml
```diff
@@ -0,0 +1,45 @@
...
+      - master
+
+  pull_request:
+    branches:
+      - "*"
+
+jobs:
+  linux:
+    name: Linux py${{ matrix.PYTHON_VERSION }} tests
+    runs-on: ubuntu-latest
```
bollwyvl, line 15: baby steps, I know, but windows (and to a lesser extent, osx) is where the real problems usually occur.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,45 @@
...
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-py-${{ matrix.PYTHON_VERSION }}-pip-${{ hashFiles('**/setup.py') }}
+      - name: Setup python
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.PYTHON_VERSION }}
+      - name: Install python dependencies
+        run: |
+          pip install setuptools pip --upgrade
+          pip install -v -e ".[test]"
```
bollwyvl, line 38: seems fine for now... though while running through it, I was a little surprised that the `jupyter_server` code paths aren't executed.

.github/workflows/tests.yml
goanpeca: OSX build machines are "expensive" So maybe on OSX we should only run  oldest and newest?

.github/workflows/tests.yml
goanpeca: 🤷

.github/workflows/tests.yml
goanpeca: Yes, will add it!

goanpeca:
Thanks @bollwyvl !

It's annoying that only after merged (unless I pushed to the repo directly) I can't really know if things are working as expected, so I would rather get this merged and I can iterate on improvements on the subsequent PRs where I can see the actual changes run :-p

.github/workflows/tests.yml
bollwyvl: too true, everything has a cost! i think as long as there is _some_ coverage on all the platforms, it's solid. another thing i've done before, when the wall-clock time isn't as important as the overall resource footprint, is do strongest and weakest indicator builds by themselves (in this case, I guess that would be windows 3.5 and and linux 3.8?) and then open up the matrix, but still set a max concurrency.

bollwyvl:
>  I can't really know if things are working 

Yeah, right? I've been opening up PRs against my fork's `master` when trying to do CI grunt work.

Even worse are the some of the events like `status` reacting to another CI's webhook: those _only_ run from `master`, so it's _really_ hard to debug them.

.github/workflows/tests.yml
bollwyvl: at which point, is 3.6 even going to tell you much more than 3.5 until EOL in september?

blink1073:
Is this ready to merge?

goanpeca:
I thinks so @blink1073 I can iterate on this further once it is merged.

blink1073:
Cool, thanks!

PR #102, opened 2020-07-28
PR: Add other OS and update py versions

### LINKED ISSUE #97: Move CI to Github actions

goanpeca:
Fixes #97

- [x] Remove travis ci
- [x] Add mac and win tests on Github actions
- [x] Update README
- [x] Add coverage


goanpeca:
@blink1073 could you add the webhook for the bot to cancel duplicate builds?

blink1073:
> @blink1073 could you add the webhook for the bot to cancel duplicate builds?

Done!

.github/workflows/tests.yml
```diff
@@ -36,10 +42,11 @@ jobs:
...
           pip install -v -e ".[test]"
+          pip install pytest-cov
       - name: Show python environment
         run: |
           python --version
           python -m pip list
       - name: Run python tests
         run: |
-          pytest . --tb=long -svv
+          pytest . --tb=long -svv --cov=jupyterlab_server
```
bollwyvl, line 52: :+1: are we gonna send coverage anywhere? If nothing else, i'm a fan of:
```
--cov-report term:skip-covered --cov-fail-under XX
```
Where XX is `whatever we have now - 1`

.github/workflows/tests.yml
```diff
@@ -11,15 +11,21 @@ on:
...
-    runs-on: ubuntu-latest
+    name: ${{ matrix.PLATFORM }} py${{ matrix.PYTHON_VERSION }}
+    runs-on: ${{ matrix.PLATFORM }}
     env:
       CI: True
       PYTHON_VERSION: ${{ matrix.PYTHON_VERSION }}
     strategy:
       fail-fast: false
       matrix:
         PYTHON_VERSION: ['3.5', '3.6', '3.8']
```
bollwyvl, line 22: again, maybe and add 3.9, and drop 3.6?

goanpeca: I tried and it did not work :-/

.github/workflows/tests.yml
goanpeca: We could 🙃

@blink1073 do we use codecov or something?

bollwyvl:
Mea culpa: the windows test timestamp fail is on me (well, and windows' terrible time resolution). Maybe we cheat and round the timestamps before comparing them? I suppose it would _still_ be possible to get occasional fails.... gah, time is why we can't have nice things.

Also might be nice to clobber the `warn` warnings with a fix, and add the traitlets warning to `filterwarnings`.

.github/workflows/tests.yml
blink1073: We aren't actively managing coverage in this org iirc but codecov is what I typically use.  I can add integration for this repo tomorrow morning.

.github/workflows/tests.yml
bollwyvl: Cool/bad! The utter fiasco that was tornado vs 3.8 makes me want to find these things out sooner rather than later. Maybe we can keep it, and just run it against linux, and expect failure? Gotta start somewhere!

goanpeca: I mean I tried

https://github.com/actions/setup-python

```
steps:
- uses: actions/checkout@v2
- uses: actions/setup-python@v2
  with:
    python-version: '3.9.0-beta.4'
```

And the install of that python did not work.

.github/workflows/tests.yml
goanpeca: Thanks!

.github/workflows/tests.yml
bollwyvl: Maybe try `3.9.0-beta.5`? The [manifest](https://github.com/actions/python-versions/blob/main/versions-manifest.json#L3) is the Source of Truth.

bollwyvl: trying this out on #103...

bollwyvl: yeah, no dice. weird.

.github/workflows/tests.yml
blink1073: We did have it set up for [`pytest-check-links`](https://github.com/jupyterlab/pytest-check-links/blob/master/.travis.yml)  I added you as a maintainer on this repo @goanpeca and created https://codecov.io/gh/jupyterlab/jupyterlab_server

blink1073: @bollwyvl you already had team access on the repo

goanpeca:
@bollwyvl so this is good to merge and you can continue on the other PR with the win fixes?

.github/workflows/tests.yml
goanpeca: Ok @blink1073 thanks!

bollwyvl:
Yeah, it's probably reasonable to introduce a new observed test fail for a PR like this. I do wonder, though, why we're not seeing better cache hits on windows: it's recompiles `pyrsistent` every time. Also the sneaky packages pip installed during the test might be disturbing the cache... perhaps `--no-cache-dir` could be used there.

bollwyvl:
Hey, i bet 3.8 is timing because of the tornado ioloop vs Proactor thing. Had to patch that in so many places :cry:

goanpeca:
>Hey, i bet 3.8 is timing because of the tornado ioloop vs Proactor thing.

Yeah, I will skip that one for now, but definitely needs a fix :-p

goanpeca:
I think this is good to go for now @bollwyvl and @blink1073 :-)

PR #96, opened 2020-07-10
PR: Add a translations handler

goanpeca:
Add a translation handler plus the utilities needed to find translations from language packs and the one bundled within packages.

This also correctly merges localization data from packages into the language packs, if the package provides a newer version.

Added tests!

PR #103, opened 2020-07-28
[wip] investigate potential json encoding issues with non-Latin characters

bollwyvl:
Now that #96 is starting to think about non-latin characters, and #102 (which this upstreams) is adding a broader test matrix, this is testing a hunch I have around non-ascii characters in schemas and settings causing issues on Windows. Would love to be proven wrong! But the fix would probably be quick (just add `encoding="utf-8"` in a few spots.


bollwyvl:
Hooray 3.5! Boo 3.8!

bollwyvl:
Ok, looking good here, but changes rather a lot of stuff... would we prefer to break this up? There's a real bug in the previous file metadata stuff i made, in addition to the focus of this PR (the unicode stuff).

goanpeca:
So maybe split into the windows fix on another PR and make this one the utf8 stuff only?

bollwyvl:
The good parts of this ended up in #107 and #108, closing.

PR #107, opened 2020-07-28
fix file mtime/ctime confusion, restore win3.8, patch event loop

bollwyvl:
- extracted from #103
- [x] fixes #106 by swapping the ctime and mtime (whoops)
- [x] adds windows vs python 3.8 test excursion and event loop hack
- [x] fixes `warn` warning
- [x] add tests stuff to `setup.cfg`
  - [x] filter known, unavoidable warnings from notebook, zmq, traitlets
  - [x] moves test args from workflow to `setup.cfg`

PR #108, opened 2020-07-28
UTF-8 all over, test with big unicode string

### LINKED ISSUE #104: HTTP 500 on windows when schemas contain non-latin characters
On a particularly gruesome schema that includes non-latin characters in a settings schema, I encountered some 500 errors on windows. [Patching those out](https://github.com/bollwyvl/jupyterlab-drawio/commit/63b15386b225232dd740e6b3541e09aa75f923b7), everything worked fine. Unfortunately I don't have logs available, but #103 (ft #102) demonstrates the condition on top of the soon-to-be-expanded test matrix.

Windows 10
conda 4.8.3
`python=3.7.8=h60c2a47_0_cpython`
`jupyterlab=2.2.0=py_0`
` jupyterlab_server=1.2.0=py_0`

bollwyvl:
- fixes #104
- builds on #107
- [x] forces `utf-8` pretty much anywhere the filesystem is involved with JSON
  - I know `utf-16`, `32`, etc. are technically allowed by the spec, but...
- [x] add tests of settings and workspace with big string of language names

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -49,7 +49,7 @@ jobs:
           python -m pip list
       - name: Run python tests
         run: |
-          pytest . --tb=long -svv --cov=jupyterlab_server --cov-report term:skip-covered --cov-fail-under 71
+          pytest --cov-fail-under 71
       - name: Upload coverage
         run: |
           codecov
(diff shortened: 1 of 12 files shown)


blink1073 (commented):
Nice, thanks!

PR #109, opened 2020-08-05
Add handling of dynamic labextensions

blink1073:


echarles:
@blink1073 Is this to handle the following issue in the log console?

Screenshot 2020-08-05 at 11 35 03

PS: I also see that the classical notebook extensions are no more loaded (they were well loaded before with a specific revision of server/nbclassic last week) - Have pinged @Zsailer also, have not yet opened any issue for that, as not completely sure where the issue/regression is coming from).

echarles:
ooh sorry, this one is for dynamic load extension... Anyway, happy to get more feedback on the 2 above points

1. extension has some null attribute at some point.
2. classical extensions not loaded anymore.

blink1073:
> extension has some null attribute at some point.

We saw this previously in https://github.com/jupyterlab/jupyterlab-module-federation/issues/42 but haven't had a chance to chase it down (we will need to before releasing final).

> classical extensions not loaded anymore.

I'll defer to @Zsailer  on this one

echarles:
> We saw this previously in jupyterlab/jupyterlab-module-federation#42 but haven't had a chance to chase it down (we will need to before releasing final).

Thx.

> I'll defer to @Zsailer on this one

Sure, I have connected with @Zsailer on that one.  It is a regression, hopefully we can fix that soon

Sorry for the noise on this PR...

blink1073:
> Sorry for the noise on this PR...

No worries, thanks for pointing these out!

PR #115, opened 2020-08-11
fix spurious logging issue

Zsailer:
This fixes the issue we've been seeing in JupyterLab with spurious logging. 

See @ellisonbg's image originally posted in jupyter/jupyter_server#272. 

image

PR #116, opened 2020-08-12
Use tilde for server requirement

blink1073:


blink1073:
I cut beta1 with this and #115

PR #117, opened 2020-08-24
app_version should be a trait

echarles:
Fixes https://github.com/jupyterlab/jupyterlab/issues/8812

`app_version` is now a trait.


blink1073:
Released as `2.0.0b2`

PR #122, opened 2020-09-01
Allow default setting overrides to be configurable in jupyter config

blink1073:
Fixes https://github.com/jupyterlab/jupyterlab/issues/8868


blink1073:
Released as `2.0.0b6`

PR #123, opened 2020-09-03
Clean up handling of config

blink1073:
Companion to https://github.com/jupyterlab/jupyterlab/pull/8950.

Moves handling of `page_config` here completely so it is dynamic.

### DIFF
--- a/jupyterlab_server/config.py
+++ b/jupyterlab_server/config.py
@@ -47,7 +47,7 @@ def get_static_page_config(app_settings_dir=None, logger=None):
     cm = ConfigManager(config_dir_name="labconfig")
     page_config = cm.get('page_config')
 
-    # TODO: remove in 4.0
+    # TODO: remove in JupyterLab 4.0
     if app_settings_dir:
         keyname = 'disabled_labextensions'
         old_page_config = pjoin(app_settings_dir, 'page_config.json')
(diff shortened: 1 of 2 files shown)


bollwyvl:
Awesome stuff here!

By dynamic, does this mean config set after starting the server would
update the defaults (with a page reload, presumably)? I'd really love a way
for nbgitpuller to work with this in binder...

blink1073:
> By dynamic, does this mean config set after starting the server would
update the defaults (with a page reload, presumably)?

Yup!

jupyterlab_server/config.py
```diff
@@ -0,0 +1,229 @@
...
+    return dynamic_exts
+
+
+def get_static_page_config(app_settings_dir=None, logger=None):
+    # Start with the deprecated `share/jupyter/lab/settings/page_config.json` data
+
+    cm = ConfigManager(config_dir_name="labconfig")
+    page_config = cm.get('page_config')
+
+    # TODO: remove in 4.0
```
blink1073, line 50: Let's instead allow passing in a `page_config` file here so it is more generic and won't need to change later.  Let's not warn here either.

bollwyvl:
Awesome. Will have to think how this could be made to work the
nbgitpuller...

blink1073:
Released as `jupyterlab_server-2.0.0b7`

Issue #113, opened 2020-08-10
Fix setting registry warnings


blink1073:
Most likely introduced in #109 

cf https://github.com/jupyterlab/jupyterlab-module-federation/issues/42

1. we need the schema field to come through in the settings response from the server
2. it doesn't exist so it's warning you about that error
3. the server doesn't require schema to exist
4. the specific plugin whose schema appears to not exist is "@jupyterlab/tooltip-extension:notebooks"
5. there is nothing obviously wrong with schema/notebooks.json in the tooltips-extension

fcollonval:
@blink1073 @afshin it seems #125 fixed this issue, isn't it?

PR #125, opened 2020-09-18
Add schemas to settings list

afshin:
The `GET` request to settings without a plugin specified should return the full list including schemas.

cf. https://github.com/jupyterlab/jupyterlab_server/issues/113
cf. https://github.com/jupyterlab/jupyterlab/issues/8958


jasongrout:
Can you release this?

PR #129, opened 2020-10-19
Improved readme / documentation

ZelphirKaltstahl:
I added some information to the readme file. This will probably help other people, when trying to make a local test setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,38 +10,13 @@ https://github.com/jupyterlab/jupyterlab_server
 `pip install jupyterlab_server`
 
 ## Usage
+
 The application author creates a JupyterLab build on their machine
 using the core JupyterLab application.  They can then serve their
 files by subclassing the `LabServerApp` with the appropriate
 configuration and creating a Python entry point that launches the app.
 
 
-## Development Install
+## Contribution
 
-``` shell
-git clone https://github.com/jupyterlab/jupyterlab_server.git
-cd jupyterlab_server
-pip install -e .
-```
-
-## Testing
-
-To create a local test setup run the following commands:
-
-``` shell
-git clone https://github.com/jupyterlab/jupyterlab_server.git
-cd jupyterlab_server
-python3 -m venv venv
-source venv/bin/activate
-pip install -e .[test]
-pytest
-```
-
-If you do not wish to create a virtual environment, you can run the following commands instead:
-
-``` shell
-git clone https://github.com/jupyterlab/jupyterlab_server.git
-cd jupyterlab_server
-pip install -e .[test]
-pytest
-```
+Please see `CONTRIBUTING.md` for details.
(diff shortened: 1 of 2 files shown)


goanpeca (changes requested):
Hi @ZelphirKaltstahl 

Thanks for the contribution. A couple of pointers:

Please create a branch from your fork and make the changes on thatt branch. You can then push and make a PR.

I am not sure, but don't think PRs from your master branch will be accepted.

Also it might be a good opportunity to use  `CONTRIBUTING.md` file instead of using the README.md


Thoughts @jupyterlab/committers ?

README.md
```diff
@@ -18,8 +18,30 @@ configuration and creating a Python entry point that launches the app.
...
+
+## Testing
+
+To create a local test setup run the following commands:
+
+``` shell
+git clone https://github.com/jupyterlab/jupyterlab_server.git
+cd jupyterlab_server
+python3 -m venv venv
+source venv/bin/activate
```
goanpeca, line 35: Since different devs could be using different setups, maybe the environment handling could be left for the user. There are many other tools (like conda) so I would rather not provide these steps, but only tell the user how to install the test dependencies with some info on using environments, defered to the specific tools (pipenv, venv condac etc.)

ZelphirKaltstahl: OK, thanks for the feedback! I will leave it neutral in terms of virtual environment options and only mention that one probably should use a tool for that.

ZelphirKaltstahl:
> Please create a branch from your fork and make the changes on thatt branch. You can then push and make a PR.

Are there specifics on what the name for such a branch should be? Perhaps `wip-improve-readme`, would that work?

README.md
goanpeca: 👍

goanpeca:
>Are there specifics on what the name for such a branch should be? 

Up to you, that is not something we really suggest or enforce. What I usually do (personal preference) is to use a prefix and a description, for example:

- `enh/add-something-cool` 
- `fix/some-bug` 
- `doc/update-something` 
- `task/improve-ci` 

Cheers!

ZelphirKaltstahl:
OK, I created another PR (https://github.com/jupyterlab/jupyterlab_server/pull/130) from a separate branch.

I was not sure what target branch to choose. Any other branch than `master` seemed even more inappropriate and I cannot create the same branch on this repository (, I think?). Hopefully it is correct, if not just tell me how to do it correctly and I will correct it. : )

goanpeca:
>I was not sure what target branch to choose. Any other branch than master

Your PR against master is ok, the thing is that the source must be a branch different fromt master on your fork. So the other PR seems fine. We can close this one :) 

Thanks again!

fcollonval:
> We can close this one :)

Closing this in light of that comment

PR #132, opened 2020-10-28
Support a metadata file in the lab extension directory

jasongrout:
Also, clean up the federated extension info a bit, so we just pull the information we need from package.json.

This is part of the strategy outlined at https://github.com/jupyterlab/jupyterlab/issues/9231


jtpio (approved):
Thanks!

jupyterlab_server/config.py
```diff
@@ -82,7 +94,7 @@ def get_page_config(labextensions_path, app_settings_dir=None, logger=None):
     federated_exts = get_federated_extensions(labextensions_path)
 
     for (ext, ext_data) in federated_exts.items():
-        if not 'jupyterlab' in ext_data or not '_build' in ext_data['jupyterlab']:
+        if not '_build' in ext_data['jupyterlab']:
```
jtpio, line 97: The `if not 'jupyterlab' in ext_data` check is not necessary anymore because federated extensions are expected to have the `jupyterlab` key in their package.json?

PR #133, opened 2020-11-06
Respect environment for page_config.json

afshin:
Respects Python environment when handling `page_config.json`

cf. https://github.com/jupyterlab/jupyterlab/issues/9240

### DIFF
--- a/jupyterlab_server/config.py
+++ b/jupyterlab_server/config.py
@@ -11,7 +11,7 @@ from os.path import join as pjoin
 import os.path as osp
 import os
 
-from jupyter_core.paths import jupyter_path
+from jupyter_core.paths import jupyter_path, jupyter_config_dir, ENV_CONFIG_PATH, SYSTEM_CONFIG_PATH
 from jupyter_server.services.config.manager import ConfigManager, recursive_update
 from traitlets import Bool, HasTraits, List, Unicode, default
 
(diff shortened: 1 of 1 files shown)


jupyterlab_server/config.py
```diff
@@ -242,3 +249,24 @@ def _default_translations_api_url(self):
...
+    Returns the string to the environment
+    Parameters
+    ----------
+    user : bool [default: False]
+        Get the user's .jupyter config directory
+    sys_prefix : bool [default: False]
+        Get sys.prefix, i.e. ~/.envs/my-env/etc/jupyter
+    """
+    if user and sys_prefix:
+        sys_prefix = False
```
jasongrout, line 265: if user is true, sys_prefix won't ever get looked at below. Why set sys_prefix to False?

jasongrout: Would it be more elegant to just have a level parameter that can be 'user', 'sys_prefix', or 'system'?

jupyterlab_server/config.py
```diff
@@ -53,10 +53,17 @@ def get_federated_extensions(labextensions_path):
     return federated_extensions
 
 
-def get_static_page_config(app_settings_dir=None, logger=None):
+def get_static_page_config(app_settings_dir=None, logger=None, all_levels=False, user=False, sys_prefix=False):
```
jasongrout, line 56: all_levels, user, and sys_prefix are mutually exclusive here, right? Is it more elegant to have a single parameter that can take three values, with a single default value?

blink1073 (approved):
Thanks!

PR #137, opened 2020-12-03
Update page_config handling

afshin:
Follow up to #133 
cf. https://github.com/jupyterlab/jupyterlab/pull/9347

* Improves layering of extension enable/disable config
* Adds handling of source extensions that disable other extensions (requires a change in [jupyterlab/jupyterlab#9347](https://github.com/jupyterlab/jupyterlab/pull/9347) to produce the metadata)
* Removes the deprecation warning for the old page config file location - save for documentation

### DIFF
--- a/jupyterlab_server/config.py
+++ b/jupyterlab_server/config.py
@@ -11,7 +11,7 @@ from os.path import join as pjoin
 import os.path as osp
 import os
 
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Thanks!

PR #140, opened 2020-12-11
Handle source extensions that are disabled

afshin:
Follow up to #137 to handle source extensions that are disabled, preventing them from disabling other extensions.


blink1073 (approved):
Thanks!

blink1073:
https://pypi.org/project/jupyterlab-server/2.0.0rc7/

PR #142, opened 2020-12-16
don't patch event loop for tornado 6.1+

bollwyvl:
tornado 6.1 works with the default event loops on all platforms now. I've been working on rooting it out of other repos, still a couple more to do.

As "the future," and having an imminent major version bump, the `jupyter_server` PR _requires_ tornado 6.1
- https://github.com/jupyter-server/jupyter_server/pull/339

As this package is already widely deployed, this uses the "just check the tornado version" approach:
- https://github.com/jupyter/notebook/pull/5907
- https://github.com/ipython/ipykernel/pull/564

Would happily do that, here, if we don't mind pining to the "bleeding edge" tornado, and remove some cruft.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,7 +40,7 @@ jobs:
           architecture: 'x64'
       - name: Install python dependencies
         run: |
-          pip install setuptools pip --upgrade --user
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
Thanks!

PR #144, opened 2020-12-19
Update jupyter_server requirement to 1.* instead of 1.1.*

jasongrout:
Update jupyter_server requirement to 1.* instead of 1.1.*

Presumably 1.2 will not be backwards compatible...


blink1073:
Thanks!

blink1073 (approved):

PR #149, opened 2021-01-08
Move open_browser to ProcessApp class

jasongrout:
This follows on the changes in ExtensionApp to move open_browser to the top-level config.


blink1073 (approved):
Thanks!

PR #150, opened 2021-01-12
Handle mathjax_url properly

afshin:
Fixes jupyterlab/jupyterlab#9605. The `static/components` url from the notebook server's `mathjax_url` config was getting prefixed with `/lab/tree` when we had a document open.


blink1073:
https://pypi.org/project/jupyterlab-server/2.1.2/

PR #154, opened 2021-01-29
Add URL and description for federated extension data

krassowski:
https://github.com/jupyterlab/jupyterlab/pull/9702 was just merged to expose the federated extension URL in the frontend, but I think that the actual URL is never set and that's why my attempt to test that PR on binder failed.

The URL (and description) is getting passed along the way here:

https://github.com/jupyterlab/jupyterlab/blob/6f4ba75604c9e6291b669119e9e42702458b6688/jupyterlab/handlers/extension_manager_handler.py#L103-L112

So it should be enough to make it work. The code for `_get_url` is extracted from:

https://github.com/jupyterlab/jupyterlab/blob/136d2ec216ebfc429a696e6ee75fee5f8ead73e2/jupyterlab/commands.py#L1404-L1410

and thinking about it now it should not be private, but exposed so that jupyterlab can re-use the same `get_url` for source extensions in the future. I will wait for feedback though.

### DIFF
--- a/jupyterlab_server/config.py
+++ b/jupyterlab_server/config.py
@@ -25,7 +25,7 @@ from .server import url_path_join as ujoin
 DEFAULT_TEMPLATE_PATH = osp.join(osp.dirname(__file__), 'templates')
 
 
-def _get_url(data):
+def get_package_url(data):
     """Get the url from the extension data
     """
     # homepage, repository  are optional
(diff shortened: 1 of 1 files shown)


krassowski:
I verified locally that it works:

Screenshot from 2021-01-30 00-23-13

The titles are non-clickable in the screenshot because I used JupyterLab 3.0.5 rather than master (so prior to https://github.com/jupyterlab/jupyterlab/pull/9702).

afshin:
This is reasonable and sounds like a good idea. May I suggest naming the function a bit more explicitly: `get_package_url()`?

krassowski:
Thanks! Renamed the function as suggested.

blink1073 (approved):
Thanks!

blink1073:
Released as https://pypi.org/project/jupyterlab-server/2.2.0/ since it adds a public API.

krassowski:
Thank you!

Issue #151, opened 2021-01-14
Installation error, WinError 206, The filename or exertion is too long.


donok1:
Hi,

Recently, I tried to install JupyterLab using pip on my new work laptop running Windows 10. While installing jupyterlab-server, the following error popped-up:

```
Installing collected packages: jupyterlab-server
ERROR: Could not install packages due to an EnvironmentError: [WinError 206] The filename or extension is too long: 'C:\\Users\\xxxx\\AppData\\Local\\Packages\\PythonSoftwareFoundation.Python.3.9_qbz5n2kfra8p0\\LocalCache\\local-packages\\Python39\\site-packages\\jupyterlab_server\\tests\\translations\\jupyterlab-language-pack-es_CO\\jupyterlab_language_pack_es_CO\\locale\\es_CO'
```

The laptop being locked up by our IT department, no chance to unlock the 260 characters path limit.

### Solution
- Clone this git
- Remove the `test` folder
- `pip install` from files

### Better solutions ?
Are these tests really necessary ?
If so, any chance to rename/re-structure them to avoid this problem ?

Thanks

hyamanieu:
This issue appears in production:
https://stackoverflow.com/questions/65749454/filenotfound-error-while-running-conda-clean-but-file-exists/66028111#66028111

Jupyterlab 3 is going to be more and more installed on Windows, this issue will pop up more often due to the default file path of Anaconda.

jasongrout:
FYI, for our automated tests we change some relevant setting in Windows: https://github.com/jupyterlab/jupyterlab/blob/a79d130576e3bac0fc73521ac7520cc6bc2b1f1a/.github/workflows/windowstests.yml#L45-L46

hyamanieu:
This was my solution, however not all users have admin rights.

PR #157, opened 2021-02-03
Reduce path length of installed package

afshin:
Fixes jupyterlab/jupyterlab_server#151 by removing the example translation packages from the distributed package. Adds a CI test against the installed tarball to make sure the test suite still passes in the installed package.


blink1073 (approved):
Thanks!

blink1073:
https://pypi.org/project/jupyterlab-server/2.1.4/

PR #162, opened 2021-02-18
Set file_url_prefix trait for opening notebooks directly

afshin:
Jupyter Server 1.4.0 exposes a new trait on `ExtensionApp`'s to let extension define their own URL prefix for opening notebooks directly. Adding it to `jupyterlab_server` here and bumping the minimum version of the `jupyter_server` dependency.


Zsailer (approved):

Zsailer:
Thanks!

blink1073:
https://pypi.org/project/jupyterlab-server/2.3.0/

Issue #168, opened 2021-03-29
API specification and documentation


jtpio:
Just like Jupyter Server with the interactive Swagger UI documenting the API: https://petstore.swagger.io/?url=https://raw.githubusercontent.com/jupyter/jupyter_server/master/jupyter_server/services/api/api.yaml#/

![image](https://user-images.githubusercontent.com/591645/112861162-d8c16c80-90b4-11eb-86e6-41b48a4c8b29.png)

It would be nice if JupyterLab Server could provide something similar to document the extra endpoints such as `settings` and `themes`:

https://github.com/jupyterlab/jupyterlab_server/blob/a82794a1f4a80ed44f5385a16cb364eb452d1ad0/jupyterlab_server/config.py#L265-L287

Maybe some static documentation might be enough as a start.

PR #169, opened 2021-03-30
Add Sphinx docs

afshin:
Add sphinx docs, with changelog and a start for the api docs.

cf. jupyterlab/jupyterlab_server#168


blink1073 (approved):
Thanks!

PR #172, opened 2021-03-31
Make LabConfig configurable and add config API docs

afshin:
Continuation of jupyterlab/jupyterlab_server#169


blink1073 (approved):
Thanks!

PR #178, opened 2021-04-19
Upgrade packaging

jtpio:
- Use `jupyter_packaging` and recommended `setup.cfg` metadata
- Use `tbump` to handle versioning and test in CI
- Add `check-manifest` support and test in CI


blink1073 (commented):
Thanks!

bollwyvl:
I guess we're shipping it, but I dislike the creep of `jupyter_packaging` with weird version constraints into runtime dependencies to get `__version__.split('.')`... also kinda looks like the runtime version spec might not be saying what we think it's saying anyhow, as `~0.9,<2` still says `>=9,<1` to my eyes...

PR #179, opened 2021-04-23
Support for overrides.json5

mlucool:
This adds support for the json5 format for setting overrides.
It will look for overrides.json then overrides.json5. Both can
be in json5 (of which JSON is a subset)

Fixes: jupyterlab/jupyterlab#10010


blink1073 (approved):
Nice, thanks!

blink1073:
I'll cut a new release with this next week.

PR #183, opened 2021-05-14
Use Jupyter Releaser

afshin:
Use [jupyter_releaser](https://github.com/jupyter-server/jupyter_releaser/) to make releases.


blink1073 (approved):
Thank you!

PR #187, opened 2021-06-15
Switch to entrypoints package

fcollonval:
Use entrypoints package to list translation entry points

Advantages:
- It is already a dependency (from the notebook server)
- it is more efficient than pkgresources
- it lists the entry points from the current sys.path => no need to run the listing in a subprocess

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,5 @@ docs/source/changelog.md
 
 # generated cli docs
 docs/source/api/app-config.rst
+
+.vscode/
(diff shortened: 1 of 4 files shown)


goanpeca (approved):
🚀

bollwyvl:
Hooray not relying on undeclared runtime `setuptools` dependency!

I would recommend adding `entrypoints` as an explicit dependency in `setup.cfg` or `.py` (wherever), as `notebook` would, along with `nbclassic`, be dropped in the already-started 4.0.0 line of lab (though `nbconvert` will continue to use it, most likely, where it's been pinned to `>=0.2.2`).

fcollonval:
Thanks for the suggestion @bollwyvl 
i updated the PR.

PR #190, opened 2021-07-21
PR: Add preferred dir option to config

goanpeca:
Following on the [discussion of adding a new trait](https://github.com/jupyter-server/jupyter_server/issues/534) `preferred-dir`

And the PR on JServer -> https://github.com/jupyter-server/jupyter_server/pull/549

This PR now exposes that option on the page config. I added both the full url and the relative path to use if found in the frontend which is using it over at https://github.com/jupyterlab/jupyterlab/pull/10667


blink1073 (approved):
Thanks!

PR #192, opened 2021-07-21
Delay list -> dict conversation to end

### LINKED ISSUE #191: Attempted config merge of list and dict clobbers dict
## Description

In the lab handler, we attempt to merge the `page_config_data` set on the webapp with the readout from the settings dir:

https://github.com/jupyterlab/jupyterlab_server/blob/52f4cf92adb8b6dc3b350ce324619a231a9003ba/jupyterlab_server/handlers.py#L124

The code that reads out data from the config file includes a part that converts dicts to lists:

https://github.com/jupyterlab/jupyterlab_server/blob/3c7a83c5212e7359fbe0955b6dda1f8024011cab/jupyterlab_server/config.py#L170-L174

However, this is run before the attempted merge with serverapp data. So, if someone has set dict data for a key (e.g. "disabledExtensions" on the serverapp data, it gets clobbered by the (potentially empty) list from the settings dir.

## Expected behavior

The conversion from dict to list happens after all merging has happened. As `recursive_update` does not merge lists, this seems the most reasonable option.

vidartf:
This should prevent dicts in webapp data from being clobbered by converted lists from the settings_dir.

Fixes #191 .

- [x] TODO: Add tests.

### DIFF
--- a/jupyterlab_server/config.py
+++ b/jupyterlab_server/config.py
@@ -97,7 +97,7 @@ def get_page_config(labextensions_path, app_settings_dir=None, logger=None):
             # Convert lists to dicts
             for key in [disabled_key, "deferredExtensions"]:
                 if key in data:
-                    data[key] = dict((key, True) for key in data[key])
+                    data[key] = dict((k, True) for k in data[key])
 
             recursive_update(page_config, data)
 
--- a/jupyterlab_server/config.py
+++ b/jupyterlab_server/config.py
@@ -167,12 +167,6 @@ def get_page_config(labextensions_path, app_settings_dir=None, logger=None):
     rollup_disabled.update(page_config.get(disabled_key, []))
     page_config[disabled_key] = rollup_disabled
 
-    # Convert dictionaries to lists to give to the front end
-    for (key, value) in page_config.items():
-
-        if isinstance(value, dict):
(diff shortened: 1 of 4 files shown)


vidartf:
@minrk This setting will interact with the hook introduced in https://github.com/jupyterlab/jupyterlab_server/pull/220 by expecting certain keys to now map to dicts instead of lists. Have you been using any such keys in hub yet?

minrk (approved):

minrk:
I haven't yet, but it shouldn't be a problem for us. Thanks for the heads up.

jtpio:
> This setting will interact with the hook introduced in #220 by expecting certain keys to now map to dicts instead of lists. 

Should this be considered a breaking change? 

After investigating the JupyterLab CI failure in https://github.com/jupyterlab/jupyterlab/pull/11744, it looks like consumers retrieving the page config with `get_page_config` now get dicts instead of lists. This is causing issues with disabled extensions in JupyterLab with  `jupyterlab_server>=2.10` when using the CLI (https://github.com/jupyterlab/jupyterlab/pull/11744#issuecomment-1001550113).

This could be fixed in lab, but would need to be able to handle both dicts and lists to allow for various combinations of `jupyterlab` and `jupyterlab_server` (`< 2.10` or `>=2.10`).

PR #188, opened 2021-07-07
Merge overrides settings if values are dicts

goanpeca:
Fixes https://github.com/jupyterlab/jupyterlab/issues/10010

When using overrides files, if a default value in a schema is a `dict` then it will update the default with the `dict` provided by the overrides file.

With `overrides.json`
```json
{
  "@jupyterlab/notebook-extension:tracker": {
      "codeCellConfig": {
         "lineNumbers": false
      }
  }
}
```

## Before this change

The resulting system default settings would be:

```json5
{
    // Notebook
    // @jupyterlab/notebook-extension:tracker
    // Notebook settings.
    // **************************************

    // Code Cell Configuration
    // The configuration for all code cells.
    "codeCellConfig": {
        "lineNumbers": false,
    },
....
```

## After this change

The resulting system default settings would be:

```json5
{
    // Notebook
    // @jupyterlab/notebook-extension:tracker
    // Notebook settings.
    // **************************************

    // Code Cell Configuration
    // The configuration for all code cells.
    "codeCellConfig": {
        "autoClosingBrackets": true,
        "cursorBlinkRate": 530,
        "fontFamily": null,
        "fontSize": null,
        "lineHeight": null,
        "lineNumbers": false,
        "lineWrap": "off",
        "matchBrackets": true,
        "readOnly": false,
        "insertSpaces": true,
        "tabSize": 4,
        "wordWrapColumn": 80,
        "rulers": [],
        "codeFolding": false,
        "lineWiseCopyCut": true
    },
....
```

PR #194, opened 2021-08-03
PR: Update new_defaults dictionary instead of copy

goanpeca:
Fixes a bug introduced on https://github.com/jupyterlab/jupyterlab_server/pull/188 which broke the overrides mechanism for defaults that ara dictionaries 😞


blink1073 (approved):
Thanks!

PR #198, opened 2021-08-16
TST: support openapi-core 0.14 SpecPath

bnavigator:
openapi-core had a severe API change: create_spec now returns a SpecPath instead of Spec.

In order to not have to pin to < 0.14 in openSUSE Tumleweed, here is a patch dealing with both APIs.

`spec['paths']` throws `openapi_core.schema.paths.exceptions.InvalidPath` in openapi-core 0.13.8, which is not importable in 0.14. The `OpenAPIError` is a common exception for the two versions.

Another possibility would be to just require ~=0.14.0 and always get a `SpecPath`.

### DIFF
--- a/jupyterlab_server/tests/utils.py
+++ b/jupyterlab_server/tests/utils.py
@@ -10,10 +10,9 @@ from urllib.parse import parse_qs, urlparse, urljoin
 from openapi_core.validation.request.datatypes import (
     RequestParameters, OpenAPIRequest
 )
-from openapi_core import create_spec
-from openapi_core.exceptions import OpenAPIError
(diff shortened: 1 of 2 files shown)


bnavigator:
https://github.com/p1c2u/openapi-core/pull/318

blink1073:
Thanks for catching this!


> Another possibility would be to just require ~=0.14.0 and always get a SpecPath

That sounds good

bnavigator:
> > Another possibility would be to just require ~=0.14.0 and always get a SpecPath
> 
> That sounds good

Updated

blink1073 (approved):
Thank you!

PR #199, opened 2021-08-17
Fix reset user settings if validation failed

fcollonval:
If user settings are not valid, `raw` is set to an empty object but `settings` is kept unchanged.

This makes `settings` equivalent to `raw` in case of validation error.


blink1073:
I was going to cut a release with #198, should I hold off a bit for this?

fcollonval:
> I was going to cut a release with #198, should I hold off a bit for this?

It is good to go on my side.

blink1073 (approved):
Thank you!

PR #202, opened 2021-08-22
Do not overwrite capitalization of region names

krassowski:
Previously the display name would return "Polish (poland)" even though the underlying babel correctly capitalizes Poland.

Fixes https://github.com/jupyterlab/jupyterlab/issues/10736#issuecomment-903280123.


blink1073 (approved):
Thank you!  I'll cut a patch release tomorrow.

PR #205, opened 2021-09-06
Translate settings schema

### LINKED ISSUE #1: Add more explicit sources of strings
### Problem

In `jupyterlab-tour`, we can now define the tours through the settings. This means the strings will not be part of the places looked for text to be translated.

> If I got it right, those are extracting by parsing the source code and taking some specified key in settings files.

I can see multiple case of string requiring to be translated that are not in the code directly (in particular in settings to customize UI elements).

Would it make sense to tell jupyterlab-translate 
- to load a specific file containing such strings
- to load any string in the settings json that have a specific annotation like `"__A to-be-translated string"`

> I have just started looking into all the machinery involved with translation. So I may definitely have missed some steps.

### LINKED ISSUE #10414: Translation of settings
### Problem

JupyterLab has the infrastructure of translating text since v3. But this works only for strings defined in the code (or title and description of settings). It would be interesting to allow the translation of strings from settings for UI built out of a data representation as:
- introduced by #10254 #10373 
- extension like jupyterlab-tour that allow tour step to be defined in settings.

### Proposed Solution

After discussion at the JLab meeting and on https://github.com/jupyterlab/jupyterlab-translate/issues/1, the path to achieve that is:

- Use jinja2 templating style on settings with functions replicating the TS translator API
- Update the settings handlers to conditional process the schema with jinja2 - default value true. But for JLab it will be false to defer the template process in the frontend where the translation is available ([nunjucks](https://mozilla.github.io/nunjucks/) will be used)
- Update the jupyterlab-translate project to extract those new strings from the settings: https://github.com/jupyterlab/jupyterlab-translate/pull/2

fcollonval:
## References

Fixes https://github.com/jupyterlab/jupyterlab/issues/10414
Fixes https://github.com/jupyterlab/jupyterlab-translate/issues/1
Closes https://github.com/jupyterlab/jupyterlab/pull/10752

| Before | After |
| --- | --- |
| image |  image |

## Code changes

The settings schemas are translated on the backend side before served to the frontend.

The code of the handlers was refactored.
The translation handler is now conditionned on the existence of settings configuration as they hold the locale source of thruth.

## User changes

Settings translatable strings are displayed translated; in particular title and description in the setting editor.

## Backward incompatible

N/A

### DIFF
--- a/jupyterlab_server/handlers.py
+++ b/jupyterlab_server/handlers.py
@@ -201,6 +201,19 @@ def add_handlers(handlers, extension_app):
         setting_path = ujoin(extension_app.settings_url, '(?P<schema_name>.+)')
         handlers.append((setting_path, SettingsHandler, settings_config))
 
+        # Handle translations.
+        ## Translations requires settings as the locale source of truth is stored in it
+        if extension_app.translations_api_url:
+            # Handle requests for the list of language packs available.
+            # Make slash optional.
+            translations_path = ujoin(extension_app.translations_api_url, '?')
+            handlers.append((translations_path, TranslationsHandler, settings_config))
+
+            # Handle requests for an individual language pack.
+            translations_lang_path = ujoin(
+                extension_app.translations_api_url, '(?P<locale>.*)')
+            handlers.append((translations_lang_path, TranslationsHandler, settings_config))
+
     # Handle saved workspaces.
     if extension_app.workspaces_dir:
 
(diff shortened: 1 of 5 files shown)


blink1073 (approved):
Thank you!

blink1073:
I'll cut this release tomorrow

krassowski:
Thank you!

PR #207, opened 2021-09-07
Fall back to `DEFAULT_LOCALE` when translation settings schema is invalid in `get_current_locale`

telamonian:
Needed to fix final CI bug for jupyterlab/jupyterlab#10929. Essentially, this PR restores the pre-#205 behavior by which a missing or malformed translation settings schema would result in `get_current_locale` returning `DEFAULT_LOCALE`. This PR also slightly improves said old behavior by producing an informative warning when this happens, instead of being silent like before.

Tested locally, fixes the problem.

pinging @fcollonval, @blink1073

### DIFF
--- a/jupyterlab_server/settings_handler.py
+++ b/jupyterlab_server/settings_handler.py
@@ -35,7 +35,14 @@ class SettingsHandler(ExtensionHandlerMixin, ExtensionHandlerJinjaMixin, SchemaH
         """Get setting(s)"""
         # Need to be update here as translator locale is not change when a new locale is put
         # from frontend
-        translator.set_locale(self.get_current_locale())
+        try:
+            locale = self.get_current_locale()
+        except web.HTTPError as e:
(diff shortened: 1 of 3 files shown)


blink1073 (approved):
Thank you!

jupyterlab_server/settings_handler.py
```diff
@@ -35,7 +35,14 @@ def get(self, schema_name=""):
         """Get setting(s)"""
         # Need to be update here as translator locale is not change when a new locale is put
         # from frontend
-        translator.set_locale(self.get_current_locale())
+        try:
+            locale = self.get_current_locale()
+        except web.HTTPError as e:
+            # fallback in case of missing (404) or misshapen (500) translation schema
+            locale = DEFAULT_LOCALE
+            'Failed loading or validating translation settings schema'
```
fcollonval, line 43: @telamonian Should this be wrapped in `self.log.warn`? Is it needed even with the handling of the error in `get_current_locale`?

PR #210, opened 2021-09-14
Clean unneeded catch

### LINKED ISSUE #209: Hanging string
@telamonian Should this be wrapped in `self.log.warn`? Is it needed to handle the error here even with the handling of the error within `get_current_locale`?

https://github.com/jupyterlab/jupyterlab_server/blob/49d9c0d8e6e24a931200c549ea83ffe9783231e0/jupyterlab_server/settings_handler.py#L38-L43

_Originally posted by @fcollonval in https://github.com/jupyterlab/jupyterlab_server/pull/207#discussion_r704169340_

fcollonval:
The error is handle in `get_current_locale` directly

Fix #209


blink1073 (approved):
Thanks!

PR #220, opened 2021-11-18
add page_config_hook

### LINKED ISSUE #219: Inject page config on a per-request basis
### Problem

When running in JupyterHub, we would like to inject some fields to page config that are based on the incoming request (e.g. based on the authenticated user, which can vary from one request to the next). However, the way page config is populated in LabHandler.get makes it hard to find a hook to add/modify fields.

### Proposed Solution

Add a hook for 'extra page config' that can be a callable, so that page config can be extended. Possibly this could be discovered via jupyter-server extensions, so that each extension could add its own page config? For us, at least some mechanism would need to to take the current request into account.

What would suffice for us:

```python
page_config = ...
# apply extra page config hook
if self.settings.get("extra_page_config", None):
    page_config = self.settings["extra_page_config"](self, copy.copy(page_config))
... # continue as before
```

but just one hook might not suffice if multiple extensions also want to do this. I'm not sure they will, though, because extensions generally have their own place to put settings. JupyterHub does need to modify the root page config itself.

### Additional context

JupyterHub's server-extension code where we would call this hook is [here][jupyterhub-singleuser]. We generally inject some things into Jupyter handler base classes. I'd like to do this via existing extension APIs as much as possible. Ultimately, I'd like to get to the point where jupyterhub-singleuser is implemented as a Jupyter Server Extension, but the necessary extension points don't exist yet.

[jupyterhub-singleuser]: https://github.com/jupyterhub/jupyterhub/blob/2.0.0rc3/jupyterhub/singleuser/mixins.py

minrk:
- move page_config to `get_page_config()` method, so it can be overridden (used in NotFoundHandler)
- support `settings["page_config_hook"]` for customizing page config on a per-request basis. Hook shoud look like:

```python
def page_config_hook(handler: RequestHandler, page_config: Dict) -> Dict:
    ...
```

closes #219

### DIFF
--- a/jupyterlab_server/handlers.py
+++ b/jupyterlab_server/handlers.py
@@ -5,6 +5,7 @@
 # Distributed under the terms of the Modified BSD License.
 import os
 from urllib.parse import urlparse
+from functools import lru_cache
 
 from tornado import template, web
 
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Thanks!

PR #222, opened 2021-12-09
pyproject.toml: clarify build system version

adamjstewart:
See https://github.com/jupyter/jupyter-packaging/issues/120

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["jupyter_packaging~=0.9,<2", "jupyter_server"]
+requires = ["jupyter_packaging~=0.9", "jupyter_server"]
 build-backend = "setuptools.build_meta"
 
 [tool.jupyter-releaser]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["jupyter_packaging~=0.9,<2", "jupyter_server"]
+requires = ["jupyter_packaging~=0.9", "jupyter_server"]
```
blink1073, line 2:
```suggestion
requires = ["jupyter_packaging>0.9,<2", "jupyter_server"]
```

blink1073 (approved):

blink1073:
Thanks!  I updated it to include 1.x support.

adamjstewart:
Was dropping 0.9 support intentional?

blink1073:
I was looking at https://github.com/jupyter/jupyter-packaging/issues/120 when I made the change.  It doesn't matter very much, since the library is only used in the isolated build environ.

PR #230, opened 2021-12-22
Automated Changelog Entry for 2.10.1 on main

blink1073:
Automated Changelog Entry for 2.10.1 on main

After merging this PR run the "Full Release" Workflow on your fork of `jupyter_releaser` with the following inputs
| Input  | Value |
| ------------- | ------------- |
| Target | jupyterlab/jupyterlab_server  |
| Branch  | main  |
| Version Spec | next |
| Since | v2.10.0 |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ github_url: 'https://github.com/jupyterlab/jupyterlab_server/blob/master/CHANGEL
 
 <!-- <START NEW CHANGELOG ENTRY> -->
 
+## 2.10.1
+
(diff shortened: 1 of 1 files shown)


bollwyvl:
Good catch, also noted on conda-forge when testing downstream:
https://github.com/conda-forge/jupyterlab_server-feedstock/pull/61

Seems like adding a downstream smoketest of lab might be good... and _maybe_ voila, but gotta stop somewhere...

blink1073:
> Seems like adding a downstream smoketest of lab might be good... and maybe voila

Both those sound good, a couple other Jupyter-maintained consumers.

PR #232, opened 2021-12-22
Add downstream test workflow

### LINKED ISSUE #231: Add downstream tests
### Problem
We are not currently testing against any downstream consumers to protect against unintentional API breakages, cf #229 

### Proposed Solution
Add downstream tests for JupyterLab and voila, cf https://github.com/jupyter-server/jupyter_server/blob/4d2f85ade21f5927c3694ff88bd6d1968bce4083/.github/workflows/downstream.yml#L30

blink1073:
Fixes #231.  voila doesn't actually depend on `jupyterlab_server`, only `jupyter_server`.

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -2,7 +2,7 @@ name: Test downstream projects
 
 on:
   push:
-    branches: "*"
+    branches: "main"
   pull_request:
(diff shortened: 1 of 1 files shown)


bollwyvl:
awesome!

PR #233, opened 2021-12-28
Fix issue where preferredPath is incorrectly calculated

mlucool:
preferredDir seems to be unsued in all of the jupyterlab codebase so this was removed as well

### DIFF
--- a/jupyterlab_server/handlers.py
+++ b/jupyterlab_server/handlers.py
@@ -83,10 +83,12 @@ class LabHandler(ExtensionHandlerJinjaMixin, ExtensionHandlerMixin, JupyterHandl
 
         server_root = os.path.normpath(os.path.expanduser(server_root))
         try:
-            page_config['preferredDir'] = self.serverapp.preferred_dir
-            page_config['preferredPath'] = self.serverapp.preferred_dir.replace(server_root, "")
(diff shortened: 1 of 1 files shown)


blink1073 (approved):
Thank you!

mlucool:
Thanks @blink1073! Is there a release planned soon? If not, I can install from this commit for the time being.

blink1073:
I'll cut a release today with this and https://github.com/jupyterlab/jupyterlab_server/pull/234

krassowski:
I think this fixed https://github.com/jupyter-server/jupyter_server/issues/647, is this right?

mlucool:
No. I think that needs to be solved here: https://github.com/jupyter-server/jupyter_server/blob/main/jupyter_server/serverapp.py#L1539.

I'm not sure what a solution looks like. Imagine:
```
> root_dir=/foo
> pwd=/foo/bar/baz
> ls $pwd
foo
bar
> preferred_dir=bar
```
What does `preferred_dir` mean here: `/foo/bar` or `/foo/bar/baz/bar`? Today it means relative to cwd like all other directories (i.e. `/foo/bar/baz/bar`). Now we want it to sometimes means relative to root_dir and sometimes to cwd.

PR #234, opened 2021-12-28
Revert "Delay list -> dict conversation to end (#192)"

blink1073:
This reverts commit 7b9c024e660a4e13a8b720fe034b7fc433239252.

Temporary fix for https://github.com/jupyterlab/jupyterlab/issues/11743

### DIFF
--- a/jupyterlab_server/config.py
+++ b/jupyterlab_server/config.py
@@ -97,7 +97,7 @@ def get_page_config(labextensions_path, app_settings_dir=None, logger=None):
             # Convert lists to dicts
             for key in [disabled_key, "deferredExtensions"]:
                 if key in data:
-                    data[key] = dict((k, True) for k in data[key])
+                    data[key] = dict((key, True) for key in data[key])
 
             recursive_update(page_config, data)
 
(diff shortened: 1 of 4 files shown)


fcollonval:
I think we can reapply #192 now that https://github.com/jupyterlab/jupyterlab/pull/11744 is merged?
cc @jtpio

jtpio:
It's not clear whether https://github.com/jupyterlab/jupyterlab_server/pull/192 should be considered as a bug fix or not. If so and assuming `get_page_config` is not really part of the public API then we could reapply it as a bug fix, now that https://github.com/jupyterlab/jupyterlab/pull/11744 handles both cases.

fcollonval:
Ok so it would be good to wait for a feedback of @vidartf before taking a decision.

> I added a note to bring that up at the next meeting.

vidartf:
Sorry, I was out for the holidays. Did that change break anything more than the test? `get_page_config` only gives the config from disk, and not the "extra" settings that the handler adds, so it has a somewhat limited scope already. The issue is ofc that it modifies the type of some returned values. Since the dict -> list conversion is mainly intended to be for the benefit of the frontend, I would argue that it makes sense to only apply this transform right before it sends the data.

#192 can be considered more of an enhancement, but it can be argued as a bug fix in that the values in `page_config_data` are not applied as you would expect.

jtpio:
> Did that change break anything more than the test?

Looks like it affected the disabling and enabling of JupyterLab extensions, based on this comment: https://github.com/jupyterlab/jupyterlab/pull/11744#issuecomment-1001550113

> #192 can be considered more of an enhancement, but it can be argued as a bug fix in that the values in `page_config_data` are not applied as you would expect.

So we could maybe reapply #192? Especially now that https://github.com/jupyterlab/jupyterlab/pull/11744 should handle both cases?

vidartf:
> So we could maybe reapply #192? 

Probably, but it might not make sense as a patch/minor release if it breaks older versions of lab.

PR #236, opened 2021-12-29
Fix workspace export without arguments

fcollonval:
Error seen in https://github.com/jupyterlab/jupyterlab/pull/11730

`jupyter lab workspaces export` should return the default workspace. This fixes it.

The default workspace id is hard coded (as it is done in JupyterLab `PageConfig.defaultWorkspace`).


blink1073 (approved):
Thanks!

PR #238, opened 2022-01-26
Also look in `static` for license files

bollwyvl:
It looks like a fair number of packages have started including `third-party-licenses.json`, but they have not been showing up in the UI/CLI.

The simple reason here being that `@jupyterlab/builder` puts them in `static` (as opposed to next to `package.json`).

This PR checks in both locations, and adds a test for the more common `static` case.


blink1073:
Kicked the check release jobs

blink1073 (approved):
Thank you!

bollwyvl:
thanks for the speedy review, and sorry for not checking it out sooner.

PR #239, opened 2022-03-13
Include Tests in Sdist But Not Wheel

blink1073:
This allows us to test on `conda-forge` with all of the tests, and use the layout recommended by `pytest`.

I'll follow this up with some CI cleanup.


setup.cfg
```diff
@@ -41,4 +41,6 @@ install_requires =
 test = codecov; ipykernel; pytest>=5.3.2; pytest-cov; jupyter_server[test]; openapi_core~=0.14.0; pytest-console-scripts; strict-rfc3339; ruamel.yaml; wheel
 
 [options.packages.find]
-exclude = ['docs*']
+exclude =
+    docs*
+    tests*
```
davidbrochart, line 46: @blink1073 I think this is responsible for [this error](https://github.com/jupyterlab/jupyterlab/runs/5571001829?check_suite_focus=true):
```
ImportError: Error importing plugin "jupyterlab_server.pytest_plugin": No module named 'jupyterlab_server.pytest_plugin'
```

blink1073: Good point, I'll make a fix

jtpio:
Looks like this makes the JupyterLab tests fail? For example with the latest commit on `master`:

https://github.com/jupyterlab/jupyterlab/runs/5583340091?check_suite_focus=true

```
E   ModuleNotFoundError: No module named 'jupyterlab_server.tests'
```

jtpio:
Ah I see there is discussion about his starting from this comment: https://github.com/jupyterlab/jupyterlab_server/pull/243#issuecomment-1069660922

davidbrochart:
I inlined `expected_http_error()` in https://github.com/jupyterlab/jupyterlab/pull/11599/commits/404fe2ed0a86457488260314672a179833ca4f8d, as @blink1073 suggested.

PR #243, opened 2022-03-16
Restore pytest_plugin

blink1073:
Follow up to #239


fcollonval (approved):
Thanks @blink1073

davidbrochart:
@blink1073 I think this is not enough, JupyterLab still needs the [test directory](https://github.com/jupyterlab/jupyterlab/blob/e1c5e0ba50c5465d7850ad2adda454f552f79ff4/jupyterlab/tests/test_build_api.py#L9), see the [CI failure](https://github.com/jupyterlab/jupyterlab/runs/5577125021?check_suite_focus=true).

blink1073:
I don't think that import makes sense for JupyterLab, it should just inline the [function](https://github.com/jupyterlab/jupyterlab_server/blob/2788986363b6884ec138ce95e4780b78cf73551a/tests/utils.py#L138).

blink1073:
We could put some of those functions into the `pytest_plugin` as an option.  But I think the `pytest_plugin` should be the only public api for testing.

jtpio:
> I don't think that import makes sense for JupyterLab, it should just inline the [function](https://github.com/jupyterlab/jupyterlab_server/blob/2788986363b6884ec138ce95e4780b78cf73551a/tests/utils.py#L138).

Just opened https://github.com/jupyterlab/jupyterlab/pull/12228 as an attempt to fix the failing CI in JupyterLab.

PR #245, opened 2022-03-17
Resolve paths in tests

blink1073:
Attempt to fix https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=476179&view=logs&jobId=656edd35-690f-5c53-9ba3-09c10d0bea97&j=656edd35-690f-5c53-9ba3-09c10d0bea97&t=e5c8ab1d-8ff9-5cae-b332-e15ae582ed2d

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -159,3 +159,5 @@ jobs:
         run: |
           cd sdist/test
           pytest -vv || pytest -vv --lf
+          # Ensure that the pytest plugin is importable.
+          python -c "from jupyterlab_server import pytest_plugin"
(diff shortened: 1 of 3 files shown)


bollwyvl:
Cool, thanks. It does still seem a bit odd that the `tests` are separated out, but not the `test_data`.

Of additional note: we're also having to do [special stuff](https://github.com/conda-forge/jupyterlab_server-feedstock/blob/main/recipe/meta.yaml#L37) to handle the the [yml file](https://github.com/jupyterlab/jupyterlab_server/blob/main/tests/utils.py#L99).

I wonder if the yml path should be hoisted to someplace in the programmatic API... not the parsed data, though, I guess, unless we pre-generated is JSON... yaml parsing in the hot path of a request (or even a server startup) seems to be an anti-feature.

PR #247, opened 2022-03-17
Clean up tests and add public openapi_spec

blink1073:
- Include a `test_utils` file that can be used as a public API
- Add `openapi_spec` and `openapi_spec_dict` conveniences

### DIFF
--- a/jupyterlab_server/__init__.py
+++ b/jupyterlab_server/__init__.py
@@ -4,6 +4,7 @@
 from .app import LabServerApp
 from .licenses_app import LicensesApp
 from .handlers import add_handlers, LabHandler, LabConfig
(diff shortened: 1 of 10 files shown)


blink1073:
@bollwyvl you had mentioned that you wanted programmatic access to the Open API spec, does this look reasonable to you?

bollwyvl:
ha, just saw this on the merge: a bit late, but I'm not entirely sold on making `ruamel.yaml` and `openapi_core` hard dependencies, and perhaps put them behind an `[openapi]` extra. 

either way, we'd still want to lazily load them, i'd imagine, as every new thing that gets referenced by `__init__.py` has a startup cost, and these aren't guaranteed to be used.

PR #250, opened 2022-03-23
Updates parameter description for notebook_starts_kernel

jweill-aws:
The parameter now maps to `preferKernel` (notebook _should_ start a kernel) instead of to `canStartKernel` (notebook _can_ start a kernel).

Related to https://github.com/jupyterlab/jupyterlab/issues/12259


blink1073 (approved):
👍🏼

PR #253, opened 2022-03-28
Make openapi optional and lazy

blink1073:


setup.cfg
```diff
@@ -49,6 +48,11 @@ test =
     strict-rfc3339
     wheel
     openapi-spec-validator<0.5
+    openapi_core>=0.14.2
+    ruamel.yaml
```
bollwyvl, line 52: it's possible to use the crusty old `%` notation to reuse components sections, though order might be important:

https://github.com/ipython/ipython/blob/master/setup.cfg#L77

PR #255, opened 2022-03-29
update URLs, metadata, extras in setup.cfg

bollwyvl:
## References
- follow on to https://github.com/jupyterlab/jupyterlab_server/pull/253#pullrequestreview-923373003
## Changes
- [x] adds URLs for docs, funding, issues, etc. to populate PyPI sidebar
- [x] refreshes trove classifiers
- [x] sorts, normalizes dependencies
- [x] uses `%()s` notation to avoid repeating package versions


blink1073 (approved):
Thanks!

PR #264, opened 2022-04-11
Switch to importlib.metadata to list entry points

fcollonval:
Alignment following https://github.com/jupyterlab/jupyterlab/pull/12293

### DIFF
--- a/jupyterlab_server/translation_utils.py
+++ b/jupyterlab_server/translation_utils.py
@@ -11,10 +11,14 @@ import re
 import sys
 import traceback
 from functools import lru_cache
+
+try:
+    from importlib.metadata import entry_points
+except ImportError:
+    from importlib_metadata import entry_points
 from typing import Dict, Pattern, Tuple
 
 import babel
-import entrypoints
 from packaging.version import parse as parse_version
 
 # Entry points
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -36,7 +36,7 @@ packages = find:
 python_requires = >=3.7
 install_requires =
     babel
-    entrypoints>=0.2.2
+    importlib_metadata>=1.4;python_version=="3.7"
```
bollwyvl, line 39: unfortunately, the situation isn't quite this simple, as the `group` argument wasn't introduced until [3.10](https://docs.python.org/3.10/library/importlib.metadata.html#entry-points) and wasn't added until 3.6 of [importlib-metadata](https://importlib-metadata.readthedocs.io/en/latest/using.html#entry-points).

So unfortunately, the safest  thing to do at this point is more like:

```ini
install_requires =
    importlib_metadata>=3.6;python_version<3.10
```

and then

```py
import sys

if sys.version_info < (3,10):
    from importlib_metadata import entry_points 
else:
    from importlib.metadata import entry_points 

# ...
    entry_points(group=JUPYTERLAB_LANGUAGEPACK_ENTRY)
```

fcollonval: Thanks @bollwyvl

jupyterlab_server/translation_utils.py
```diff
@@ -14,9 +14,13 @@
 from typing import Dict, Pattern, Tuple
 
 import babel
-import entrypoints
 from packaging.version import parse as parse_version
 
+if sys.version_info < (3, 10):
```
bollwyvl, line 19: Might even be worth linking to that thing about why this complexity is required.

blink1073 (approved):
Thanks!

jupyterlab_server/translation_utils.py
fcollonval: Thanks

PR #271, opened 2022-05-03
Switch to flit build backend

blink1073:


pyproject.toml
```diff
@@ -1,14 +1,79 @@
...
+Source = "https://github.com/jupyterlab/jupyterlab_server"
+Tracker = "https://github.com/jupyterlab/jupyterlab_server/issues"
+
+[project.optional-dependencies]
+openapi = [
+  "openapi_core>=0.14.2",
+  "ruamel.yaml",
+]
+test = [
+  "openapi_core>=0.14.2",
```
bollwyvl, line 56: might as well sort

pyproject.toml
```diff
@@ -1,14 +1,79 @@
...
+  "jupyter_server[test]",
+  "openapi-spec-validator<0.5",
+  "pytest-console-scripts",
+  "pytest-cov",
+  "pytest>=5.3.2",
+  "strict-rfc3339"
+]
+
+[project.license]
+file = 'LICENSE'
```
bollwyvl, line 69: maybe normalize quotes

pyproject.toml
blink1073: Done!

pyproject.toml
blink1073: Done!

bollwyvl:
it's so pretty i love it

blink1073:
> it's so pretty i love it

Yeah it is a yummy diff

PR #282, opened 2022-06-09
Add `jupyter-lsp` to dependencies list

trungleduc:
This is a part of the ongoing effort (https://github.com/jupyterlab/team-compass/issues/148, https://github.com/jupyter-server/team-compass/issues/30) to support LSP in core JupyterLab.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,7 @@ dependencies = [
   "jinja2>=3.0.3",
   "json5",
   "jsonschema>=3.0.1",
+  "jupyter-lsp",
   "jupyter_server>=1.8,<2",
   "packaging",
   "requests",


fcollonval:
@trungleduc is this required? As far as I can tell it is only required in `jupyterlab` package with https://github.com/jupyterlab/jupyterlab/pull/12534, isn't it?

fcollonval:
I would rather keep it as direct dependency of `jupyterlab` package than this one to be consistent. Otherwise this adds an unused dependency that will bring maintenance troubles - for instance in the future if `jupyter-lsp` API changes, you will need to update `jupyterlab` and `jupyterlab-server` (the latter just to bump the version constraint).

fcollonval:
So I'm closing this one.

PR #302, opened 2022-09-08
raise jupyter_server pin to include 2.x

Zsailer:
This raises the ceiling on JupyterLab Server's jupyter_server dependency.

This enables folks to test Jupyter Server 2.0 with JupyterLab 4.0.

cc @fcollonval


fcollonval (approved):
Thanks @Zsailer 

I'm gonna do a minor rc release with this one

PR #334, opened 2022-11-08
Increase coverage

blink1073:
Goal is to hit 80%

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,7 +29,7 @@ jobs:
           - os: ubuntu-latest
             python-version: "pypy-3.7"
           - os: ubuntu-latest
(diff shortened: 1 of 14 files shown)


bollwyvl:
> Base: 65.11% // Head: 80.45% // Increases project coverage by +15.34% tada

well done, old chum! 

I applaud the goal, and do wonder if we can get some more on-going bang for the coverage buck with e.g.  [schemathesis](https://schemathesis.readthedocs.io/en/stable/)... nothing wrong with cranking up the specificity of the API description as well. These might want to be gated behind a flag, as they can be... long. Especially once you use them for a while.

fcollonval (approved):
Thanks @blink1073

PR #343, opened 2022-12-08
Adopt ruff and address lint

blink1073:
cf https://github.com/jupyter-server/jupyter_server/pull/1114


bollwyvl:
Just as I see you pushing, but apropos of nothing:

```
Typing :: Typed
```

is now a trove classifier. Maybe we start using around the house?

PR #353, opened 2022-12-14
Exclude test_data from wheel

### LINKED ISSUE #352: test_data in the main package
I'd like to ask what is the point of having the test_data folder as a part of the main package. If I, as a user, install jupyterlab-server, I get this folder but it does not seem to be used anywhere outside tests that are not part of the main package.

fcollonval:
Fixes #352


fcollonval:
Closing because we need those data for the pytest plugin (I forget about it...)

https://github.com/jupyterlab/jupyterlab_server/blob/264256ec4eb0a4b7e22a6a60a294081adc806146/jupyterlab_server/pytest_plugin.py#L90

PR #360, opened 2022-12-24
Add support for newer openapi spec

### LINKED ISSUE #359: Update to latest `openapi-spec-validator`
### Problem

There is some bad metadata in the upstreams:

- https://github.com/p1c2u/openapi-spec-validator/issues/173

### Proposed Solution

Go ahead and update to, and only support the lastest versions of, `openapi-core`, `openapi-spec-validator`. 

### Additional context

- https://github.com/conda-forge/jupyterlab_server-feedstock/pull/83
- a `pip check` somewhere _might_ help catch this, but as they are `[extras]`....
- `py.typed` is provided in the more recent versions, which should help things
- we'll likely end up with an OpenAPI wrapper for tornado, which would be virtuous to upstream, and reuse in other places

blink1073:
Fixes #359


blink1073:
I'll release this next week.

pyproject.toml
```diff
@@ -63,24 +63,24 @@ docs = [
...
   "ruamel.yaml",
 ]
 test = [
   "codecov",
   "ipykernel",
   "pytest-jupyter[server]>=0.6",
-  # openapi_core 0.15.0 alpha is not working
-  "openapi_core~=0.14.2",
-  "openapi-spec-validator<0.6",
+  "openapi_core>=0.16.1",
```
bollwyvl, line 73: minor nit, but perhaps we keep it DRYer with `jupyterlab_server[openapi]`

lresende:
Why this is forcing `jsonschema>=4.17.3` while the [openapi](https://github.com/globality-corp/openapi/blob/develop/setup.py) requires only `jsonschema[format]>=2.6.0` ? 

If nothing specific, could we revert back to a more flexible requirement here `jsonschema>=3.0.1`, particularly as packages will usually cap on the next major release and we suddenly moved from 3.0.1 as the minimum to 4.17.13 the latest? 

@blink1073 @marthacryan

blink1073:
`openapi-spec-validator` requires it.

PR #366, opened 2023-01-05
Don't change language environment variable

fcollonval:
This inverts the logic to by default use the default OS locale to set JupyterLab language (if the associated language pack is installed).

It uses the class API of Python `gettext` instead of the global API as recommended by the Python documentation.


blink1073 (approved):
Thanks!

blink1073:
I'll make a minor release tomorrow.

fcollonval:
Thanks for fixing the CI and the release @blink1073

PR #373, opened 2023-02-13
Reduce minimumn requirement for jsonschema to version 3.2.0

lresende:
- Currently various external dependencies caps jsonschema on the next major release, and we jumped from requiring jsonschema 2.6.0 to 4.17.3 on #360 and that is causing a lot of incompatibility in the ecosystem.

- This reduces the requirement to 3.2.0 which is also the minimun required by jupyter-events


blink1073:
4.0.1 has been available since Sep 2021, I'm failing to see the issue.

lresende:
> 4.0.1 has been available since Sep 2021, I'm failing to see the issue.

There isn't an issue per see. What I am trying to say is that, if we don't explicitly depend on a specific functionality from 4.x it would be good to be a little more flexible on the cap as we are seeing a lot of issues downstream as dependencies are still stuck on defining `<=4.0`.

A concrete example is that I can't deploy Elyra with the latest JupyterLab as kfp requires jsonschema<=4.0. I see different issues at work as well, etc.

blink1073:
We do require it, for https://github.com/p1c2u/openapi-spec-validator/blob/e798c5af68bd7d2efc1484a876e26bdb82d245fd/pyproject.toml#L47.  I'd understand if it was a more recent release, but it is 18 months old and it is unlikely the `jsonschema` 3.x is getting any critical bugfixes.

lresende:
> We do require it
We define the minimum to be 4.0, but are we leveraging capabilities that are not available in an older version? In other words, would the code stop compiling/working if we deploy with 3.x? 

Also, this is mostly giving flexibility to users. With this change, I could deploy an application together with JsonSchema version 3.x if a transient dependency requires this... or I could indeed use the latest JsonSchema if I am deploying lab with other dependencies that are more up to-date. 

This change, In no way, forces a deployment with 3.x, but mostly enables it if the dependency gets resolved to it due to transient dependencies.

lresende:
An example of similar flexibility:
https://github.com/jupyter/jupyter_events/blob/c0ea8b03428de6e36652a781f9d60f69a98f7cc8/pyproject.toml#L25-L35

lresende:
Closing this in favor of working with the original cause of this behavior
https://github.com/kubeflow/pipelines/pull/8855

Issue #375, opened 2023-02-22
jupyterlab_server fails when allowed_extensions_uris is set


rmoe:
## Description
Starting a server with this option enabled results in the following traceback:

```python
    Traceback (most recent call last):
      File "/opt/conda/lib/python3.10/site-packages/jupyter_server/extension/manager.py", line 356, in load_extension
        extension.load_all_points(self.serverapp)
      File "/opt/conda/lib/python3.10/site-packages/jupyter_server/extension/manager.py", line 228, in load_all_points
        return [self.load_point(point_name, serverapp) for point_name in self.extension_points]
      File "/opt/conda/lib/python3.10/site-packages/jupyter_server/extension/manager.py", line 228, in <listcomp>
        return [self.load_point(point_name, serverapp) for point_name in self.extension_points]
      File "/opt/conda/lib/python3.10/site-packages/jupyter_server/extension/manager.py", line 219, in load_point
        return point.load(serverapp)
      File "/opt/conda/lib/python3.10/site-packages/jupyter_server/extension/manager.py", line 147, in load
        return loader(serverapp)
      File "/opt/conda/lib/python3.10/site-packages/jupyter_server/extension/application.py", line 466, in _load_jupyter_server_extension
        extension.initialize()
      File "/opt/conda/lib/python3.10/site-packages/jupyterlab/labapp.py", line 781, in initialize
        super().initialize()
      File "/opt/conda/lib/python3.10/site-packages/jupyter_server/extension/application.py", line 430, in initialize
        self._prepare_handlers()
      File "/opt/conda/lib/python3.10/site-packages/jupyter_server/extension/application.py", line 320, in _prepare_handlers
        self.initialize_handlers()
      File "/opt/conda/lib/python3.10/site-packages/jupyterlab/labapp.py", line 777, in initialize_handlers
        super().initialize_handlers()
      File "/opt/conda/lib/python3.10/site-packages/jupyterlab_server/app.py", line 107, in initialize_handlers
        add_handlers(self.handlers, self)
      File "/opt/conda/lib/python3.10/site-packages/jupyterlab_server/handlers.py", line 295, in add_handlers
        ListingsHandler.pc = ioloop.PeriodicCallback(
      File "/opt/conda/lib/python3.10/site-packages/tornado/ioloop.py", line 886, in __init__
        if callback_time <= 0:
    TypeError: '<=' not supported between instances of 'tuple' and 'int'
```
## Reproduce

[...]

PR #376, opened 2023-02-23
Change periodic callback time from tuple to int (#376

rmoe:
Closes: #375


rmoe:
@blink1073  is there something else I need to do to get these tests to pass? I can run `hatch run test:nowarn` locally and it passes. Are these tests expected to fail here?

blink1073:
Thanks @rmoe!  The pre-release failure is unrelated, the openapi libraries keep changing APIs on us.

Issue #377, opened 2023-02-24
page_config.json should support json5


mlucool:
### Problem

Similar to overrides, `page_config.json` should also support json5 so that comments can be added.

### Proposed Solution

All json configs should always support json5. Maybe we add a common codepath to enforce this? I.e. instead of doing something like [this](https://github.com/jupyterlab/jupyterlab_server/blob/eb37a21d1deaf449074da2d087b57a773e26fdaa/jupyterlab_server/config.py#L91) we refactor [this](https://github.com/jupyterlab/jupyterlab_server/blob/eb37a21d1deaf449074da2d087b57a773e26fdaa/jupyterlab_server/settings_utils.py#L291) to be reused.

peytondmurray:
I'm happy to take this on! Refactoring now to allow json5 everywhere.

PR #378, opened 2023-02-24
[WIP] Add JSON5 support across the code base

peytondmurray:
This PR adds support for JSON5 across the code base, closing #377. Everywhere there was a call to `json.load` has been replaced with a json5-compatible loader.

The `load_json` function had to be added to a separate file because it is needed in both `settings_utils.py` and `translation_utils.py`, and if I put it in `settings_utils.py` there's a circular dependency.

### DIFF
--- a/jupyterlab_server/config.py
+++ b/jupyterlab_server/config.py
@@ -3,7 +3,6 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-import json
 import os.path as osp
 from glob import iglob
 from itertools import chain
@@ -13,6 +12,8 @@ from jupyter_core.paths import SYSTEM_CONFIG_PATH, jupyter_config_dir, jupyter_p
 from jupyter_server.services.config.manager import ConfigManager, recursive_update
 from jupyter_server.utils import url_path_join as ujoin
 from traitlets import Bool, HasTraits, List, Unicode, default
+
+from .json_utils import load_json
 
 # -----------------------------------------------------------------------------
 # Module globals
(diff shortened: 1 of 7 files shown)


bollwyvl:
this seems... hasty.

In a core jupyter product, a proper, standards-based interchange format should be consumed and generated strictly as possible, even if that's not what VSCode does. 

I think any non-standard formats should require a clear indicator (e.g. `.json5`). 

Ideally, we'd be using a more performant, but still conformance-tested, JSON parser than stdlib, as _hundreds_ of files are parsed at startup and during runtime, and I can't imagine `json5` is doing any favors to that aspect, security, etc.

If anything, `.toml` has emerged as something jupyter servers can support natively, soon enough, with `tomllib`, with a suitable backport available for `tomli`.

peytondmurray:
> I think any non-standard formats should require a clear indicator (e.g. `.json5`).

That makes sense to me - and if people feel like this is worth doing, we can definitely make that happen.

> Ideally, we'd be using a more performant, but still conformance-tested, JSON parser than stdlib, as _hundreds_ of files are parsed at startup and during runtime, and I can't imagine `json5` is doing any favors to that aspect, security, etc.

You're absolutely right, in fact that's the first item in the Known Issues section of the pyjson5 readme.
 
> If anything, `.toml` has emerged as something jupyter servers can support natively, soon enough, with `tomllib`, with a suitable backport available for `tomli`.

I'm all in favor of migrating configurations to toml, especially given `tomllib`'s addition to the standard library. I'm not sure what the path forward from json-based configuration would look like - it seems like it would make sense to support both formats for some time before json gets deprecated.

@bollwyvl Sorry if this draft was cause for alarm, I really didn't intend it to be. I was interested in making this a WIP space for seeing what breaks when I do this, what the performance implications are, making space for a dialogue, and so on. I can understand if this is a maintenance burden, and I'd be happy to close it if so.

peytondmurray:
Okay, after a discussion with @mlucool we've decided to make a JEP to standardize config formats. With that in mind I'll close this PR out to move the discussion to the forthcoming JEP. In the meantime, and if others won't object, I'll make a new PR to add json5 support just for `page_config.json`. There's already a precedent for this here - see https://github.com/jupyterlab/jupyterlab_server/blob/d114d4931d533c11570372abc27eed6fd2aca399/jupyterlab_server/settings_utils.py#L71.

Issue #380, opened 2023-02-28
Improving page load time


mlucool:
### Problem
As I was debugging https://github.com/jupyter-server/jupyter_server/issues/312, I found that certain lab APIs used for pageload are called many times without caching. This leads to a large increase in pageload time in at least some circumstances, namely if the FS is slow, this doubles my page load time for a bare bones jupyterlab from ~3s to ~6s. 

This can be traced to `get_settings`. Using `%%time`:
```
%%time
x = get_settings(app_settings_dir=app_settings_dir, labextensions_path=labextensions_path, overrides= {}, schema_name='', schemas_dir=schemas_dir, settings_dir=settings_dir);
```
With a settings_dir that is fast (SSD), I see:
> CPU times: user 40.2 ms, sys: 4.28 ms, total: 44.5 ms
> Wall time: 69.3 ms

With my setting_dir on a slower file system (NFS):
> CPU times: user 323 ms, sys: 9.66 ms, total: 333 ms
> Wall time: 401 ms

Why this matters is this API is called 17 times to load the page leading to up to (401-70)*17=5.6s more time to load the page (this is a max, in practice some settings are faster to load than `schema_name=''`)
![image](https://user-images.githubusercontent.com/1813603/222002429-e6420fe7-e431-4e6e-956e-3042a0018de1.png)

Digging deeper via pyinsturment, the cause for me is mostly in [_get_user_settings](https://github.com/jupyterlab/jupyterlab_server/blob/main/jupyterlab_server/settings_utils.py#L71):
![image](https://user-images.githubusercontent.com/1813603/222002856-67b61137-76c6-4010-9b16-f23324f7d92a.png)
Surprisingly this is blaming loads, which does not match my mental model, but [json5.loads](https://github.com/jupyterlab/jupyterlab_server/blob/main/jupyterlab_server/settings_utils.py#L92) is not known to be fast (and often it is passing in a schema that never changes as its the default and not from userspace).

The fact that there are 17 calls adds a bit more pain as I _think_ there is some HTTP1.1 head of line blocking and/or tornado being single threaded blocking going on which forces this to be serialized. 

Creating reproducers here is a bit hard, but adding this [here ](https://github.com/jupyterlab/jupyterlab_server/blob/main/jupyterlab_server/settings_utils.py#L85)seems to simulate it ok-ish (I'm not sure it's realistic):
```
    import time
    time.sleep(0.1)

    if os.path.exists(path):
        stat = os.stat(path)
[...]

PR #385, opened 2023-03-15
Optional argument to get only the schema names

brichet:
Jupyterlab requests the settings to get only the schema ids [here](https://github.com/jupyterlab/jupyterlab/blob/a396badbb6a3c840bdc088b22463fbcb42376bbe/packages/apputils-extension/src/settingsplugin.ts#L41).
This request takes some time as it send all the schema and settings content.

This PR ads an optional argument to the `api/settings` request to send only the ids of the schema without parsing the files.

Related to https://github.com/jupyterlab/jupyterlab_server/issues/380

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -111,7 +111,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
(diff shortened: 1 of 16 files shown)


fcollonval (approved):
Thanks @brichet

PR #388, opened 2023-03-21
Add json5 support for page_config.json

### LINKED ISSUE #377: page_config.json should support json5

peytondmurray:
Closes #377.

Following on from [this issue](https://github.com/jupyterlab/jupyterlab_server/issues/377), the discussion [here](https://github.com/jupyterlab/jupyterlab_server/pull/378), and a discussion from the [jupyter_server team meeting](https://github.com/jupyter-server/team-compass/issues/45#issuecomment-1471227025), this PR adds json5 support for `page_config.json`. `json5` is _only_ used to load the config if a `page_config.json5` is found to avoid incurring a performance overhead for users who just want to use `json`. An existing test was modified to provide coverage for this.

### DIFF
--- a/jupyterlab_server/config.py
+++ b/jupyterlab_server/config.py
@@ -9,7 +9,7 @@ from glob import iglob
 from itertools import chain
 from os.path import join as pjoin
 
-import json5
+import json5  # type:ignore
 from jupyter_core.paths import SYSTEM_CONFIG_PATH, jupyter_config_dir, jupyter_path
 from jupyter_server.services.config.manager import ConfigManager, recursive_update
 from jupyter_server.utils import url_path_join as ujoin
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
Thank you!

PR #394, opened 2023-04-21
Set Cache-control immutable for static files

brichet:
As mentioned in https://github.com/jupyterlab/jupyterlab/issues/10273 and https://github.com/jupyterlab/jupyterlab/issues/14124, JupyterLab now uses hashes for the javascript assets.

There is probably no need to keep the `Cache-control=no-cache` anymore, as the hash should ensure that the cached file is the expected one.

~~This PR adds a handler for static files and uses it in tornado settings.~~

Thanks to @fcollonval and @davidbrochart too.

**EDIT:**
This PR uses the web app settings key included in https://github.com/jupyter-server/jupyter_server/pull/1268

### DIFF
--- a/jupyterlab_server/app.py
+++ b/jupyterlab_server/app.py
@@ -3,11 +3,15 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
+from glob import glob
+from os.path import relpath
+
 from jupyter_server.extension.application import ExtensionApp, ExtensionAppJinjaMixin
+from jupyter_server.utils import url_path_join as ujoin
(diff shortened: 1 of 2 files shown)


brichet:
Some tests using the *performance insight* of Chrome, with the times to get the *Largest Contentful Paint* depending on the throttler set up:

|                   | 30 Mbit/s, 2 ms latency | 10 Mbit/s, 10 ms latency |
| :----            |                 :----:               |                 :----:                 | 
| immutable | ~2.6 s                            | ~2.8 s                              |
| no-cache   | ~2.8 s                            | ~3.3 s                              |

fcollonval (approved):
Thanks @brichet

PR #417, opened 2023-10-31
Add repository info when getting federated extensions

nbowditch-einblick:
## Code changes

- Adds missing `repository` field to values returned by `get_federated_extensions` if it's available in relevant package.json (and if it contains a "url" field)

### DIFF
--- a/jupyterlab_server/config.py
+++ b/jupyterlab_server/config.py
@@ -60,6 +60,11 @@ def get_federated_extensions(labextensions_path: list[str]) -> dict[str, Any]:
                     dependencies=pkgdata.get("dependencies", dict()),
                     jupyterlab=pkgdata.get("jupyterlab", dict()),
                 )
+
+                # Add repository info if available
+                if "repository" in pkgdata and "url" in pkgdata.get("repository", {}):
+                    data["repository"] = dict(url=pkgdata.get("repository").get("url"))
+
                 install_path = osp.join(osp.dirname(ext_path), "install.json")
                 if osp.exists(install_path):
                     with open(install_path, encoding="utf-8") as fid:


nbowditch-einblick:
> Thanks @nbowditch-einblick
> 
> Is the `url` defined [there](https://github.com/jupyterlab/jupyterlab_server/blob/471d6903bebfbca19f86c4d810b8195a2949a71a/jupyterlab_server/config.py#L56) not properly set up and used [there](https://github.com/jupyterlab/jupyterlab/blob/91c9e52b09791847989e5ef132651cbb6e74d6bf/jupyterlab/extensions/manager.py#L557)?

Thanks for the response!

The  `url` field is indeed correctly set with the value of `homepage` in our **package.json** file. However, the iissue is that we have a non-github url for our `homepage`. In our case, we have a landing page we're going to host on our company's site to use as a homepage, but I've also seen other extensions use their pypi page as a homepage. It seems reasonable that these urls might be different, i.e., that the package's homepage might be different from its github/repo url.

In the [logic on the front-end side](https://github.com/jupyterlab/jupyterlab/blob/fc124f46c3c1705eeb4b96638624f1acc6e6912b/packages/extensionmanager/src/widget.tsx#L26), it should be able to use either a `homepage` or `repository_url` field to get  a github user uri:

```ts
function getExtensionGitHubUser(entry: IEntry) {
  if (
    entry.homepage_url &&
    entry.homepage_url.startsWith('https://github.com/')
  ) {
    return entry.homepage_url.split('/')[3];
  } else if (
    entry.repository_url &&
    entry.repository_url.startsWith('https://github.com/')
  ) {
    return entry.repository_url.split('/')[3];
  }
  return null;
}
```

The object being read in that front-end logic from there is constructed in **manager.py** in jupyterlab's repo, [here](https://github.com/jupyterlab/jupyterlab/blob/fc124f46c3c1705eeb4b96638624f1acc6e6912b/jupyterlab/extensions/manager.py#L554). The object being read in **manager.py** is what is returned from the `get_federated_extensions` function that I changed.

The `url` field seems a bit ambiguous, I'm not sure I agree with how it can pull from either the `homepage` or `repository.url` in a given **package.json** file. It seems useful to have distinct definitions for each url. That being said, I don't what other logic uses this as a dependency; it might be more appropriate in other contexts.

fcollonval (approved):
Thanks @nbowditch-einblick  for this and the very detailed information.

PR #424, opened 2023-11-12
Fix link-rot for Swagger view of OpenAPI spec

nopdotcom:
Was the spec ever in `/docs/`? Also, the primary branch name changed, and GitHub's attitude about `master -> main` is "hey, we made a new preference and put out a press release; all our work here is done."

### DIFF
--- a/jupyterlab_server/rest-api.yml
+++ b/jupyterlab_server/rest-api.yml
@@ -1,4 +1,4 @@
-# see me at: http://petstore.swagger.io/?url=https://raw.githubusercontent.com/jupyterlab/jupyterlab_server/master/docs/rest-api.yml#/default
+# see me at: https://petstore.swagger.io/?url=https://raw.githubusercontent.com/jupyterlab/jupyterlab_server/main/jupyterlab_server/rest-api.yml#/default
 openapi: "3.0.3"
 info:
   title: JupyterLab Server


nopdotcom:
It's frustrating to be still finding these.

GitHub seemed to be under immense time pressure to ship a "name of the first branch" preference item, but the engineering effort stopped when GitHub could put out that press release saying "`main` is now the default branch for new users."

Support for the many existing GitHub customers making this transition was not a priority, even when low-effort work could make the bookkeeping less bumpy. Some repos have weird default branches; they could be ignored. But there are many, many projects which started default-`master` and switched to default-`main`, and GitHub wasted all of our time by making individual projects deal with this without help. But again, once their press release was out, their social responsibility seemed to be over.

By default for a transitioned project, the web server at `raw.githubusercontent.com` should serve a permanent redirect from the `/organization/proj/master/` prefix to the `/organization/proj/main/` tree. This is safe if `master` does not exist and `main` does. (For projects which had to transition back, supporting the opposite redirect under the opposite condition seems safe.) 

Ideally, repo owners could control the deprecation process, ranging from an even safer “don’t redirect, serve directly under the old, wrong name” to “return ‘410 Gone’ with an explanatory HTML error page.”

(Personally, I think all of this effort could have been much better spent; like many systems, Git’s `master` referred to a version of data — think “remastered recording” — rather than a role. “Master as role” is why protocols like SPI and Ethernet teaming need a deeper renaming. Ironically, Git was the first major version control system designed so that *there is no privileged repository*: the default branch in my personal Linux kernel Git repo is neither upstream nor downstream from Linus’s — or rather, the roles aren’t fixed forever at creation time. While I find this deeply satisfying politically, it’s also just flat-out a better design.)

fcollonval (approved):
Thanks @nopdotcom

PR #444, opened 2024-03-11
Ignore pageconfig file if JSON is zero-length

### LINKED ISSUE #443: Zero length page_request.json throws exception
## Description

A zero-length page_request.json will cause connections to the server to fail with error 500.

## Reproduce

`echo "" > ~/.jupyterlab/labconfig/page_config.json`

## Expected behavior

Normal connection to notebook (with a warning emitted about invalid json, perhaps).

## Context

- Operating System and version: Centos 7, Alma8, Alma9
- Browser and version: Firefox, Chrome
- JupyterLab version: 3.6.7 

We've seen this happen at least twice: a user has filled their disk/quota, and `page_config.json` gets written out as a zero-length file.

holzman:
Fixes #443

### DIFF
--- a/jupyterlab_server/config.py
+++ b/jupyterlab_server/config.py
@@ -126,19 +126,15 @@ def get_page_config(
             pjoin(app_settings_dir, "page_config.json"),
         ]
         for path in config_paths:
-            if osp.exists(path):
-                try:
-                    data = load_config(path)
-                    # Convert lists to dicts
-                    for key in [disabled_key, "deferredExtensions"]:
-                        if key in data:
-                            data[key] = {key: True for key in data[key]}
+            if osp.exists(path) and osp.getsize(path):
+                data = load_config(path)
+                # Convert lists to dicts
+                for key in [disabled_key, "deferredExtensions"]:
+                    if key in data:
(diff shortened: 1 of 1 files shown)


fcollonval (approved):
Thanks @holzman

krassowski:
Can pageconfig include security related options?  I think yes. This I think it's better to crash than to allow the server to start with the options inactive. Of course that should be with a better error message

manics:
Regardless of if it's a security issue, from a user perspective it can be very annoying/surprising to have a config file silently ignored (effectively) due to a syntax error. There may be log messages, but people generally don't look at them unless there's a problem, which may not be immediately apparent since JupyterLab will start.

Same for https://github.com/jupyter-server/jupyter_server/pull/1404

In the motivating issue https://github.com/jupyter-server/jupyter_server/issues/1403 the problem is a lack of disk space, which is going to cause many more problems than an invalid empty config file.

manics:
Sounds reasonable to me!

fcollonval:
@krassowski does the final version is ok for you?

krassowski (approved):
Thanks!

PR #445, opened 2024-03-20
For v2 extensions parse entrypoint data

AshokChoudhary11:
This change will make entrypoint data available for jupyterlab application.  
It will be consumed for loading the plugins in a no-code way.  Refer https://hackmd.io/DBlDBDszSVSjYK9ieg1mZA

### DIFF
--- a/jupyterlab_server/config.py
+++ b/jupyterlab_server/config.py
@@ -163,6 +163,8 @@ def get_page_config(
             extension["mimeExtension"] = extbuild["mimeExtension"]
         if "style" in extbuild:
             extension["style"] = extbuild["style"]
+        if "version" in extbuild and "entrypoint" in extbuild and extbuild["version"] == '2':
+            extension['entrypoints'] = extbuild["entrypoint"]
         extensions.append(extension)
 
         # If there is disabledExtensions metadata, consume it.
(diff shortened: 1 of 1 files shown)


jupyterlab_server/config.py
```diff
@@ -163,6 +163,8 @@ def get_page_config(
             extension["mimeExtension"] = extbuild["mimeExtension"]
         if "style" in extbuild:
             extension["style"] = extbuild["style"]
+        if "version" in extbuild and "entrypoint" in extbuild and extbuild["version"] == "2":
+            extension["entrypoints"] = extbuild["entrypoint"]
```
fcollonval, line 167: You could simplify that:

```suggestion
        # FIXME @experimental for plugin with no-code entrypoints.
        extension["entrypoints"] = extbuild.get("entrypoints")
```

The results will be `extension["entrypoints"] is None` if it is missing in the package.json file.

> I propose to use `entrypoints` plural too in the package.json instead of singular.

fcollonval (changes requested):
Thanks @AshokChoudhary11 

I proposed a simplification.

Could you write down some prose in the PR description to explain why you are proposing this changes and what it adds?

fcollonval (approved):
Thanks @AshokChoudhary11

PR #450, opened 2024-04-23
Move ENV_CONFIG_PATH import back into _get_config_manager

divyansshhh:
We move the import for ENV_CONFIG_PATH back into the _get_config_manager function so that patches applied in jupyterlab tests are correctly reflected

### DIFF
--- a/jupyterlab_server/config.py
+++ b/jupyterlab_server/config.py
@@ -13,7 +13,7 @@ from os.path import join as pjoin
 from typing import Any
 
 import json5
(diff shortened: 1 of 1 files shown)


jupyterlab_server/config.py
```diff
@@ -370,6 +370,8 @@ def _get_config_manager(level: str, include_higher_levels: bool = False) -> Conf
     """Get the location of config files for the current context
     Returns the string to the environment
     """
+    from jupyter_core.paths import ENV_CONFIG_PATH
```
krassowski, line 373: Restoring the comment from previous version too:

```suggestion
    # Delayed import since this gets monkey-patched in tests
    from jupyter_core.paths import ENV_CONFIG_PATH
```

PR #452, opened 2024-05-22
Load overrides once, rather than in each settings handler

krassowski:
### References

Fixes https://github.com/jupyterlab/jupyterlab/issues/16355

### Code changes

Previously the `overrides.json` were loaded from disk for each instance of the handler. Now the loading takes part once prior to handler intiialization.

### User-facing changes

When `overrides.json` is present, JupyterLab loads faster.

### Breaking changes

None


fcollonval (approved):
Thanks @krassowski

PR #456, opened 2024-07-14
Fix handling language packs for pseudo language

fcollonval:
Fix handling the language pack for the pseudo-language allowing in-app translation.

Needed by https://github.com/jupyterlab-contrib/in-context-translation


krassowski (approved):
Thank you!

> I would consider whether `PSEUDO_LANGUAGE` should be accepted by `is_valid_locale`, but current implementation works fine too.

PR #457, opened 2024-07-19
Add `@web.authenticated` for `ListingsHandler`'s `GET` method

krassowski:
The listing handler was leaking the configuration (`blocked_extensions_uris`, `allowed_extensions_uris`, `blocked_extensions`, `blocked_extensions`). This should not need an advisory so opening PR directly.


jtpio (approved):
Thanks!

PR #459, opened 2024-08-06
Add allow_setup_custom_env_variables flag to page config

AnastasiaSliusar:
This PR is the part of solution of the feature where users can set up custom env variables for a kernel which has not been run yet.
UI is on https://github.com/AnastasiaSliusar/jupyterlab/pull/2

Code changes
This PR includes `allow_setup_custom_env_variables` flag in page config. This flag is used to detect whether a user should see UI for setup custom env variables or not.


krassowski:
@AnastasiaSliusar what is the status here? Is it still being pursued?

PR #463, opened 2025-02-10
Fix typo in default for template directory traitlet (`template_dir` → `templates_dir`)

### LINKED ISSUE #462: `template_dir` or `templates_dir`?
https://github.com/jupyterlab/jupyterlab_server/blob/bca8978ffd44093bc4e7c790d8ee064d2f2b492a/jupyterlab_server/config.py#L252-L254

matches

https://github.com/jupyterlab/jupyterlab_server/blob/bca8978ffd44093bc4e7c790d8ee064d2f2b492a/jupyterlab_server/config.py#L321-L323

but

https://github.com/jupyterlab/jupyterlab_server/blob/bca8978ffd44093bc4e7c790d8ee064d2f2b492a/jupyterlab_server/config.py#L241

does not

https://github.com/jupyterlab/jupyterlab_server/blob/bca8978ffd44093bc4e7c790d8ee064d2f2b492a/jupyterlab_server/config.py#L317-L319

krassowski:
Fixes #462

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -20,7 +20,7 @@ permissions:
 jobs:
   analyze:
     name: Analyze
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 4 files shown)


fcollonval (commented):
Good catch @krassowski 

I bump the version of Python in #464 in the hope to avoid the failing tests due to warnings.

But it seems the typing issue in the lint job and the translation tests need anyway more work.

PR #464, opened 2025-02-12
Upgrade Python version in tests

fcollonval:
Tests are failing in #463 due to warnings. As 3.8 is not supported, this is an easy attempt around the test failure.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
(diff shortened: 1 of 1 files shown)


krassowski:
Closing as the min version was bumped in https://github.com/jupyterlab/jupyterlab_server/pull/472. We should probably still add 3.13 or maybe even 3.14 (rc3)

Darshan808:
> Closing as the min version was bumped in https://github.com/jupyterlab/jupyterlab_server/pull/472. We should probably still add 3.13 or maybe even 3.14 (rc3)

Just opened https://github.com/jupyterlab/jupyterlab_server/pull/473

PR #469, opened 2025-09-10
Add documentation for `page_config_hook`

jtpio:
`page_config_hook` was added in https://github.com/jupyterlab/jupyterlab_server/pull/220.

However it is not easily discoverable.

So this PR adds it to the documentation, so searching for 'hook' can help find it more easily.


jtpio:
> Rebasing here should help now that 470 is closed 🤞

Done.

krassowski (approved):
Thanks!

PR #471, opened 2025-09-24
Add `untracked_message_types` to Page Config

Darshan808:
### Description

Based on our discussion in the PR, we're exposing `untracked_message_types` to the Page Config.
This change helps the frontend by giving it a clear way to identify which messages shouldn't update the kernel's status. As a result, we can solve the issue of unnecessary kernel status updates (https://github.com/jupyterlab/jupyterlab/issues/14578)

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -20,7 +20,7 @@ permissions:
 jobs:
   analyze:
     name: Analyze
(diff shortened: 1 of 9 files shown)


krassowski:
^ The test failures look related

Darshan808:
#472 Fixes the CI.

krassowski (approved):

PR #472, opened 2025-09-26
Fix CI Failures

### LINKED ISSUE #470: Fix CI
## Description

CI appears to be broken for new PRs.

As noticed in https://github.com/jupyterlab/jupyterlab_server/pull/469

https://github.com/jupyterlab/jupyterlab_server/actions/runs/17605522377/job/50015411832?pr=469

Image

## Reproduce

Happens with new PRs.

## Expected behavior

CI should pass

## Context

We likely need to drop Python 3.8 and modernize the CI a bit.

Darshan808:
## Fixes #470

This PR addresses CI failures that have been affecting recent pull requests.

#### 1. Python 3.8 Hatch Installation Failure
```
TypeError: 'type' object is not subscriptable
Error installing hatch.
```
- The CI matrix was using `"3.8"`, but modern versions of `hatch` and its dependencies require Python 3.9+ 
- Changed `"3.8"` to `"3.9"` in the test matrix

#### 2. Unused Type Ignore Comment
Removed the unused `# type: ignore` comment that is no longer needed.

#### 3. Deprecated Ubuntu Runner Causing Job Stalls

- Some jobs were stalling with `Waiting for a runner to pick up this job...` because it was using the deprecated `ubuntu-20.04` runner.
- Updated CodeQL workflow to use ubuntu-latest (currently ubuntu-22.04) for reliable runner availability.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -20,7 +20,7 @@ permissions:
 jobs:
   analyze:
     name: Analyze
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
 
     strategy:
       fail-fast: false
--- a/.github/workflows/fix-license-header.yml
+++ b/.github/workflows/fix-license-header.yml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   header-license-fix:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
 
     permissions:
       contents: write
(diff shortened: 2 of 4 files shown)


Darshan808:
The license workflow is still using `ubuntu-20.04` because it uses `pull_request_target`, which runs the workflow from main branch. 

Once this PR is merged, future PRs will use the updated `ubuntu-latest` version.

krassowski (approved):
Thank you @Darshan808!

PR #473, opened 2025-10-08
Bump Python max version to 3.14 in CI

Darshan808:
Python 3.14 has been released, so this PR updates the CI configuration to include it as the new maximum supported version.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ["3.9", "3.12"]
+        python-version: ["3.9", "3.14"]
         include:
           - os: windows-latest
             python-version: "3.9"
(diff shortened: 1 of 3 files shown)


krassowski:
Based on logs it looks like this will require updating:

https://github.com/jupyterlab/jupyterlab_server/blob/1acc30bc46d71e1e79866271c31d9af75ffb6303/jupyterlab_server/test_utils.py#L164-L182

Darshan808:
Based on logs, it seems the error originates from [jupyter_core_code](https://github.com/jupyter/jupyter_core/blob/db8d93d31ad09c0d5e33ed6d91c73c5d87b2447f/jupyter_core/utils/__init__.py#L184-L185) using [asyncio.WindowsSelectorEventLoopPolicy](https://docs.python.org/3.14/library/asyncio-policy.html#asyncio.WindowsSelectorEventLoopPolicy). According to the [Python 3.14 deprecations](https://docs.python.org/3.14/deprecations/index.html), this policy has been deprecated.

Darshan808:
I see this deprecation being tracked in https://github.com/jupyter/jupyter_core/pull/445

Darshan808:
I think there's still a bug in jupyter_core which is causing tests to fail.
Opened https://github.com/jupyter/jupyter_core/pull/449

Darshan808:
`jupyter_core` was released with the bug fix. But the log now says `jupyter_server` also needs some fixes to support Python 3.14 on windows.
Opened https://github.com/jupyter-server/jupyter_server/issues/1565

PR #474, opened 2026-01-31
Drop use of undocumented `list2cmdline` in favor of `shlex.join`

krassowski:
`shlex.join` was introduced in Python 3.8 and since we require 3.8 or newer we can just use it. `list2cmdline` is Windows-only and not formally documented.

Noticed when looking at https://github.com/jupyterlab/jupyter-builder/issues/48

### DIFF
--- a/jupyterlab_server/process.py
+++ b/jupyterlab_server/process.py
@@ -8,6 +8,7 @@ import atexit
 import logging
 import os
 import re
+import shlex
 import signal
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


krassowski:
```
    File "/home/runner/test_venv/downstream_test/jupyterlab/jupyterlab/commands.py", line 40, in <module>
      from jupyterlab_server.process import Process, WatchHelper, list2cmdline, which
  ImportError: cannot import name 'list2cmdline' from 'jupyterlab_server.process' (/home/runner/test_venv/lib/python3.12/site-packages/jupyterlab_server/process.py)
```

Ah, it looks like this is used downstream here:

https://github.com/jupyterlab/jupyterlab/blob/8559db5726988e516c0347f2a4a1b3e4c941902a/jupyterlab/commands.py#L97

 I will restore it as an alias for now.

Darshan808 (commented):
This looks good.
I think moving process to `jupyter_builder` is also a good idea.

PR #479, opened 2026-09-15
Fix develop extensions feature with tornado 6.5.9+

krassowski:
Depends on https://github.com/jupyter-server/jupyter_server/pull/1703


andrii-i (approved):
Thanks @krassowski, this looks good. Tested locally with jupyter_server 2.21.1 + tornado 6.5.10: a symlinked labextension 403s on main and serves correctly with this branch. Happy to contribute the repro as a regression test if useful.

krassowski:
Noting that CI did not run fully, re-enabling manually now:

image

PR #480, opened 2026-09-15
Add regression test for serving symlinked labextensions

andrii-i:
Adds a regression test for the symlinked labextension (`jupyter labextension develop`) serving fixed by #479: on tornado >= 6.5.9 the request 403s without that fix and passes with it.

Depends on #479.


krassowski (approved):
Thank you!

PR #481, opened 2026-09-15
Fix CI, bump min Python version to 3.10

krassowski:
Addresses most failures seen in #479


andrii-i (approved):
Thanks @krassowski, this looks good.

